PR #57, opened 2021-12-17
make leaderboard show specific run

### LINKED ISSUE #56: Anonymous runs do not show up on leaderboard
@bricehalder I think you caused this in #37 cause runs without a `user_id` don't go to the frontend anymore. Can you take a look?

bricehalder:
fixes #56 

Now always shows run when given `run_id` of prompt. So this will work for anonymous runs as well as repeat runs from registered users


bricehalder:
example of repeat run
Screen Shot 2021-12-17 at 12 50 29 AM

example of anonymous run
Screen Shot 2021-12-17 at 12 51 30 AM

bricehalder:
not 100% sure if the SQL is the best/cleanest, open to alternatives lol

dqian3 (approved):
Seems good enough to me, a little messy, but yeah not much we can do except do the filtering on the frontend.

PR #60, opened 2022-01-03
Integrated Scraper, added manage functions

mliu59:


dqian3 (changes requested):
Some initial notes
- What is the purpose of `apis/pathFinder/backup`? Do we need it?
- One worry that I have: If the search takes too long, it may prevent the server from handling other requests until it is finished. I will have to do more research on whether that is true in production, but regardless, it's probably a good idea to try it out
- In terms of code organization, it would be better to move scraper out from `apis/pathFinder` and put them in a top-level module. I'd rather have `apis` only handle requests/responses (and currently database requests though I want to change that). It would be similar to how the achievements are currently organized.

mliu59:
- Backup isnt technically needed, just wanted to keep those in case the wiki dump parser doesnt work in the future for some reason. I'll keep a local copy and remove from branch
- From my testing so far, all requests to the scraper finishes within 5 seconds, regardless of length. Only worry is if some weird article isnt being accounted for in the database, or if some edges are wrong, but those haven't come up at all yet
- sure, can make that change

dqian3:
For the scraper taking a while, I still want to investigate it a bit more, just to make sure it's ok. There's a few ways that we could change things would definitely settle my mind a bit though, in order of increasing difficulty

1. See if our deployment has the ability to handle multiple requests at once, and increase that if so
2. Run the scraper in a different thread
3. Write a separate service for the scraper, and put it on a different server, so it doesn't take up the same pool of resources. 

For all options 2 and 3 too, something that might be better is that when you start the thread/send the request, then you can respond to the client. Then the client can either poll or manually refresh to see the results. Then, the flask server should be free to handle other requests while the other thread/service is going

dqian3:
I'm still gonna go through the code too

mliu59:
Added a single thread timeout "checking" wrapper in util. Originally wanted to use multithreading (signal/multiprocessing), but flask keeps raising "Working outside of application context." error.

dqian3 (approved):

PR #65, opened 2022-01-05
Ratings

dqian3:


scripts/createDB.py
```diff
@@ -26,6 +26,19 @@
 );
 ''')
 
+# Add info for better calculations
+TABLES['ratings']=(
+'''
+CREATE TABLE IF NOT EXISTS `ratings` (
```
bricehalder, line 32: I think we talked about supporting rating history right? I think we can have the fields be 
`user_id`
`rating`
`date` 

and add a new entry whenever a user plays a rated run. 

We can scrap `num_rounds` since we can just query for the number of entries of a user in the ratings table. Lmk what you think - one con would definitely be more storage used.

To make computation on ratings easier, maybe we can also maintain a table `current_ratings` that have the most up-to-date rating for a user. Its fields can just be 2 FKs
`user_id`
`rating_id`

So update both tables on a new rating update.

scripts/calculate_rating.py
```diff
@@ -0,0 +1,151 @@
...
+import pymysql
+import json
+from pymysql.cursors import DictCursor
+
+import itertools
+import pprint
+
+USER_QUERY = 'SELECT user_id, username from users'
+
+# TODO technically if a user has the same start time, they'll show up twice, this could be an issue
```
bricehalder, line 16: Like if a user starts two runs at the same time? Not sure about the scenario here. If it's a matter of uniqueness, perhaps another GROUP BY after the final ORDER BY might work?

scripts/calculate_rating.py
```diff
@@ -0,0 +1,151 @@
...
+def _rating_for_seed(users, round, u, desired_seed):
+    lo = 1
+    hi = 8000
+
+    while (lo < hi):
+        mid = (lo + hi) // 2
+
+        mid_seed = 1
+
+        # TODO should we exclude ourselves?
```
bricehalder, line 71: I think the person setting the prompt of the day probably shouldn't play it - honor system is fine lol

bricehalder (commented):
Didn't look too much into the Vue changes, but lmk what your thoughts are on changing this to support rating history - shouldn't be too bad and I can also take a crack at it if you want

scripts/createDB.py
dqian3: Yeah, just having a date attached to the rating is what I was thinking as well. However, I don't know if we can do that with our current ranking system: if someone goes back and plays an old ranked prompt it'll affect everyone. So I think I'm gonna try and implement the prompt of the day stuff first, and just use those for the ratings.

scripts/calculate_rating.py
dqian3: Yeah if they do the exact time, or somehow submit a duplicate, I think maybe let's put this in as an issue and figure something out later

scripts/calculate_rating.py
dqian3: Oh this is something with the rating system, whether your own rating factors into calculating your expected place

PR #66, opened 2022-01-05
profile.js fixes

mliu59:


mliu59:
Branch can be closed after merge

static/js/profile.js
```diff
@@ -34,24 +34,25 @@ var vm = new Vue({
 
 function update_data(runs, user) {
             
-    vm.totalratedruns = runs.length;
-    vm.email = user[0]['email'];
-    if (user[0]['email_confirmed']) {
+    vm.totalratedruns = runs['total_prompts'];
+    vm.email = user['email'];
```
dqian3, line 38: What happens when the user[`email`] isn't in the response? I think this would error, but I'd rather just not show the email field in the first place

static/js/profile.js
```diff
@@ -34,24 +34,25 @@ var vm = new Vue({
...
         vm.emailverified = "Yes";
     } else {
         vm.emailverified = "No";
     }
     
-    vm.user_name = user[0]['username'];
+    vm.user_name = user['username'];
 }
 
 async function get_data(usern) {
```
dqian3, line 48: This should be wrapped in `window.onload` rather than get called from the script. I think because the script is at the end it might be fine, but I'm not sure. Other option is to just switch it to vue

dqian3: While we're at it, let's change `un` argument in the profile endpoint to like `profile_name`

dqian3 (changes requested):

dqian3:
Can you change the name of the branch too?

Also the link in the navbar doesn't work

image

PR #68, opened 2022-01-07
Profile.js fix

mliu59:
rename from scraper.js-fix, includes a number of small fixes


dqian3 (approved):
Looks good to me, I think with vue we might not need to window add event listener, but it should be fine.

In the future, we would want to put options to edit the profile if a user is looking at their own profile. (i.e. send email confirm request, reset password, chagne username, etc.)

PR #71, opened 2022-01-09
Prompts of the day

dqian3:


mliu59 (approved):
Looks good to me. Were we able to test the rating system with any placeholder/existing data?
couple of small changes we can probably make to front end, but may be best to open another front end branch

>

dqian3:
> Looks good to me. Were we able to test the rating system with any placeholder/existing data? couple of small changes we can probably make to front end, but may be best to open another front end branch
> 
> >

Yeah we were able to test it, I'll probably comment it out for now though, until we get some rated prompts in

PR #92, opened 2022-01-21
Prompt page updates: vue, pagination, minor graph fixes

mliu59:


static/js/play.js
```diff
@@ -176,16 +178,16 @@ async function finish() {
 function hideElements() {
     
     var hide = ["reference","mw-editsection","reflist","portal","refbegin", "sidebar", "authority-control", "external", "sistersitebox"]
-    for(i=0; i<hide.length; i++) {
+    for(var i=0; i<hide.length; i++) {
```
dqian3, line 181: small thing, but if you're going through and fixing these, you should use `let`/`const` instead of `var`

static/js/prompt.js
```diff
@@ -216,6 +244,104 @@ function rgbToHex(r, g, b) {
     return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
 }
 
+var app = new Vue({
```
dqian3, line 247: Is this paginated only on the frontend? I think it should also be on the backend, which is a bit more tricky, but might be fine for now

dqian3 (changes requested):

static/js/play.js
mliu59: done, made those changes at first because I had to change the js type to module

dqian3:
It seems to me that some of the styling for the play page got messed up? 

Also I helped merge master changes in, I think I did it right, but not 100% sure, lmk if you see anything

dqian3 (changes requested):
see above comment

mliu59:
Fixed a couple of issues with rendering on play and prompt

dqian3 (changes requested):
Found another issue
image

I think there's still some more undeclared vars hanging around. This has definitely convinced me to move to a real JS framework.  For now, maybe try running our code through here https://jshint.com/ to clean it up

dqian3 (approved):

PR #93, opened 2022-01-23
Use single server data module instead of templates everywhere

dqian3:


templates/users/reset_password.html
```diff
@@ -2,8 +2,11 @@
 
 {% block head %}
   
-<script defer>
+<script type="module">
+import { serverData } from "./static/js/modules/getServerData.js"
```
bricehalder, line 6: I think this should be `serverData.js` instead of `getServerData.js`

templates/base.html
```diff
@@ -2,32 +2,35 @@
...
 
-        {% if username %}
-        <script defer>
-            const user_id='{{user_id}}';
-        </script>
-        {% else %}
-        <script defer>
-            const user_id = null;
+        <script type="module">
+            import {setServerData} from "/static/js/modules/serverData.js";
```
bricehalder, line 13: Might need to be `./static/.../` ? Not sure

bricehalder (changes requested):
I think this will be a useful refactor, nice. Just some minor things with the file paths

templates/base.html
dqian3: no, because we need an absolute path, the relative one changes depending on which page we visited,

templates/users/reset_password.html
dqian3: will fix

PR #94, opened 2022-01-25
Disable search shortcuts

maggarwal99:
- Instead of checking search shortcuts to catch potential cheaters, disabling it entirely for smoother game experience.
- Handling cases for more OS, keyboards, and browsers. Current check only works for Chrome, Safari and Ctrl keyboard layouts. 
- Tested with {OS: [Linux, Windows, MacOS], keyboards: [ctrl, meta], browsers: [Chrome, Safari, Firefox]}. NOTE: this is not robust to key remappings.
- Caveat: users will still be able to search the page by manually going into Settings -> Find. This is difficult to prevent. One potential solution is to obfuscate the entire page by randomly embedding invisible span elements.

### DIFF
--- a/static/js/play.js
+++ b/static/js/play.js
@@ -311,6 +311,12 @@ function checkForFind(e) {
     }
 }
 
+function disableFind(e) {
+    console.log(e);
+    if ([114, 191, 222].includes(e.keyCode) || ((e.ctrlKey || e.metaKey) && e.keyCode == 70)) { 
+        e.preventDefault();
+    }
+}
 
 window.addEventListener("load", async function() {
     const response = await fetch("/api/prompts/" + prompt_id);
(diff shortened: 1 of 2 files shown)


bricehalder:
Since we're removing all uses of checkForFind, I think we should remove the function itself as well. Thanks @maggarwal99 !!!

maggarwal99:
(y) I was gonna cleanup the code after getting this reviewed.

static/js/play.js
```diff
@@ -311,6 +311,12 @@ function checkForFind(e) {
     }
 }
 
+function disableFind(e) {
+    console.log(e);
+    if ([114, 191, 222].includes(e.keyCode) || ((e.ctrlKey || e.metaKey) && e.keyCode == 70)) { 
+        e.preventDefault();
```
bricehalder, line 317:
```
        if (ctrlfwarnings == true) {
            app.$data.finished = true;
            app.$data.caught = true;
        }
```
Need to add this part in from checkForFind if we want to end people's runs here. Although we can make an argument for just disabling it without ending the run as well.

bricehalder (changes requested):

static/js/play.js
dqian3: Personally, I think we should go the middle path, and just like flash a warning or something rather than end the run. If we want to scare people more, we can say something like 'this will be recorded'

bricehalder (approved):

PR #114, opened 2022-01-26
Make hyperlinks non-findable

maggarwal99:
- Embeds hyperlink characters into `div` blocks to make them non-searchable in case players bypass search block.


dqian3:
small thing, but the link hover styling doesn't work after this change?

dqian3 (changes requested):
See the above comment

dqian3 (approved):

PR #126, opened 2022-01-28
Scraper cleanup, fixed typing syntax error

mliu59:


dqian3:
I removed the timeout stuff, cause it was causing errors with our querying. For now I think we can leave this unmerged, but I'll work off of this branch for the scraper-task integration

dqian3 (approved):

PR #147, opened 2022-01-31
Immediate start

melodymhsu:
+ Added button to immediately start a run during countdown
- Removed extraneous "/' in about & acknowledgements page

### DIFF
--- a/static/js/play.js
+++ b/static/js/play.js
@@ -13,6 +13,7 @@ let app = new Vue({
         countdown: 8,
         finished: false,
         started: false,
+        immediateStart: false,
         activeTip: "",
         path:[],
         finalTime:"",
@@ -35,6 +36,11 @@ let app = new Vue({
 
         home: function (event) {
             window.location.replace("/");
+        },
+
+        startNow: function (event) {
+            app.$data.started = true;
+            app.$data.immediateStart = true;
         }
 
     }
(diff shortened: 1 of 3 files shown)


static/js/play.js
```diff
@@ -35,6 +36,11 @@ let app = new Vue({
 
         home: function (event) {
             window.location.replace("/");
+        },
+
+        startNow: function (event) {
+            app.$data.started = true;
```
dqian3, line 42: I know other places in this file use `app.$data`, but you can also use `this` instead, which is more concise

static/js/play.js
```diff
@@ -35,6 +36,11 @@ let app = new Vue({
...
+            app.$data.immediateStart = true;
```
dqian3, line 43: I know other places in this file use `app.$data`, but you can also use `this` instead, which is more concise

static/js/play.js
dqian3, line 42: I think you need to do 2 more things here:
1. Stop the countdown timer so startTime doesn't get reset (see [here](https://developer.mozilla.org/en-US/docs/Web/API/clearInterval)
2. Set the startTime here. I'm not sure why it's not inside the vue instance, but no need to worry about that now

static/js/play.js
dqian3: The other places use `app.$data` because they're not members of the vue component it seems

static/js/play.js
dqian3: nvm about 2, it seems that we start when we load. However we don't need to set it below either even for quick start

static/js/play.js
```diff
@@ -285,8 +291,9 @@ function countdownOnLoad(start, end) {
         if (distance < -1000) {
             clearInterval(x);
             app.$data.started = true;
-            
-            startTime = Date.now();
+            if (!app.$data.immediateStart) {
```
dqian3, line 294: don't need to set startTime at all, since it is set in loadPath initially

dqian3 (changes requested):

dqian3 (approved):
Actually, it seems like I've misread the entirety of `play.js`. This solution works for now, but we need to clean it up for sure.

Issue #24, opened 2021-12-06
Account for Load times in game.js


dqian3:
Find some way to pause timer/subtract offset to the time

bricehalder:
I thought we don't want the loading time to be included in the run time? Not sure what you mean @mliu59

PR #168, opened 2022-02-09
Clean up leaderboard logic for first run

### LINKED ISSUE #155: Loading screen time is counted in runs
Right now we load the page as soon as the user enters, and it submits the run start time. So the time that the user waits for the countdown counts against them.

We could do a quick fix, but I think overhauling the play.js page is the right way to go, and we can do things like #24 and #54

### LINKED ISSUE #167: Bold users run in leaderboard after finish run

bricehalder:
#155 was mostly handled in #159, this PR is just some cleanup so we can close #155 and #24 .

- Log run in database with NULL start time upon clicking a prompt
-> Need to run `ALTER TABLE runs MODIFY start_time TIMESTAMP(3) NULL;` in prod db
- Leaderboard query for first run changed to use earliest `run_id` of a user instead of earliest `start_time`

Verified user run appears on leaderboard only if user is logged in and completed the run on their first try

Also fixes #167

### DIFF
--- a/apis/prompts.py
+++ b/apis/prompts.py
@@ -197,13 +197,14 @@ def get_prompt_leaderboard(id, run_id):
     SELECT run_id, path, runs.user_id, username, TIMESTAMPDIFF(MICROSECOND, runs.start_time, runs.end_time) AS run_time
     FROM runs
     JOIN (
-            SELECT users.user_id, username, MIN(start_time) AS first_run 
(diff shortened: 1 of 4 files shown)


dqian3:
My only comment on this is idk if tying run_id to the submitted time is the best idea?

dqian3 (approved):

PR #183, opened 2022-02-21
Play.js refactoring

mliu59:
Some basic refactoring. Got rid of many redundant variables and functions. 

play.js initializes and populates a Vue object with prompt information, submits the run as an empty run, then passes the object to a js module to handle the rest of the game.


mliu59:
Some things that would be nice to have: 
- Further separate page loading and game logic (loadPage())
- allow the module playGame() function return game results when finished, then play.js would handle submitting runs at the beginning and end as well
- Make some functions that are gamemode specific more modular
- instead of ignoring some links (in handleWikipediaLink(e)), we just completely filter those out in parseAndCleanPage()
-

mliu59:
Countdown experiences a really long lag at the beginning for some reason

mliu59:
merge conflicts resolved. Also solved the issue that causes graphs on prompt page to render weirdly if any of the rendered runs has the starting article visited again

PR #185, opened 2022-02-22
added username filter

mliu59:
Checks for vulgar words when users are registering (based on an online list)

List is not directly stored as a text file but rather as an encrypted file, so we won't be directly storing said list lol


dqian3 (changes requested):
Consider using a hash function instead of symmetric encryption, that makes more sense to me.

mliu59:
How would one use hash functions to check if a hashed string is a substring of a normal string? Symmetric encryption allows for data recovery, so its pretty straightforward there, but I'm not sure about hashing. 

Ex. check if "hello" contains the substring "hel". "hel" is only available as a hashed string. "hello".hash() does not contain "hel".hash(). How to work around this?
image

mliu59:
I guess i could brute force and hash every possible substring in the checked string to see if it matches some hash, but is there another way to work around this?

dqian3:
Closing this for now, I think the better approach would be to setup better moderation tools and do a better job of verifying accounts.

PR #192, opened 2022-02-25
fix bug in passwords.py

bricehalder:
python not c++

### DIFF
--- a/scripts/create_db.py
+++ b/scripts/create_db.py
@@ -26,7 +26,7 @@ def create_database(db_name, recreate=False, test_config=None):
 
     with conn.cursor() as cursor:
         if (recreate):
-            drop_db_query = f'DROP DATABASE {db_name}'
+            drop_db_query = f'DROP DATABASE IF EXISTS {db_name}'
             cursor.execute(drop_db_query)
 
         create_db_query = f'CREATE DATABASE IF NOT EXISTS {db_name}'


wikispeedruns/auth/passwords.py
```diff
@@ -25,7 +25,7 @@ def check_password(user: dict, password: str) -> bool:
     id = user["user_id"]
     hash = user["hash"]
 
-    if (user["is_old_hash"]):
+    if "is_old_hash" in user and user["is_old_hash"]:
```
dqian3, line 28: Why would this not be in user? It should always be queried?

dqian3 (changes requested):
W

wikispeedruns/auth/passwords.py
bricehalder: Screen Shot 2022-02-24 at 7 31 32 PM

this was the error I was trying to fix - not sure if it's just an issue locally or not.

I was trying to log into a user that was created before the recent password changes

dqian3: can you add the full stack trace?

bricehalder: Screen Shot 2022-02-24 at 9 18 57 PM

Repro:
on recent pull of main branch, no local changes
attempt to login to account on local instance

bricehalder: nvm, solution is to re-run createDB script

PR #194, opened 2022-02-25
Profile fix

bricehalder:
Fills profile page with relevant user data

Screen Shot 2022-02-24 at 8 38 11 PM

### DIFF
--- a/apis/profiles.py
+++ b/apis/profiles.py
@@ -28,10 +28,10 @@ def get_user_info(username):
     return result, 200
 
 
-@profile_api.get("/<username>/totals")
+@profile_api.get("/<username>/stats")
 def get_total_stats(username):
     '''
-    Get the total number of prompts for a user
+    Get aggregate statistics for a user
     TODO cache this?
     '''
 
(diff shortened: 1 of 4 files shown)


apis/profiles.py
```diff
@@ -13,11 +13,20 @@ def get_user_info(username):
...
-    if (session.get("username") == username or session.get("admin")):
-        query = "SELECT `username`, `email`, `email_confirmed`, `admin` FROM `users` WHERE `username`=%s"
-    else:
-        query = "SELECT `username`, `email_confirmed`, `admin` FROM `users` WHERE `username`=%s"
 
+    admin_query_string = 'admin,' if session.get("username") == username or session.get("admin") else ''
+    query = f'''
+    SELECT
+        username,
+        email,
```
dqian3, line 21: We shouldn't expose people's emails

dqian3: that was the point of only querying for email if the user it itself.

dqian3 (changes requested):

apis/profiles.py
bricehalder: Oh oops, yeah this was unintentional good catch

bricehalder: fixed

dqian3 (approved):

PR #203, opened 2022-02-27
added catch to filter runs with invalid paths

mliu59:


apis/sprints.py
```diff
@@ -185,10 +185,15 @@ def get_prompt_leaderboard(id, run_id):
...
+        filtered = [x for x in results if not x['path'] == None]
+
+        for run in filtered:
+            try:
+                run['path'] = json.loads(run['path'])
+            except TypeError as e:
+                print(e)
+                print(run)
         
-        resp["leaderboard"] = results
```
dqian3, line 192: This shouldn't be needed actually? Cause it's always valid json, unless maybe you have some old data

PR #209, opened 2022-02-27
Alert user to login to play rated/prompt of the day

arpansahoo:


dqian3 (changes requested):
This looks great! Thanks for the contribution.

Could you also make sure that when a user goes to the `play` page that they get a similar alert and get redirected? That's what we had initially but it got lost in some other changes.

It may seem redundant, but it's needed if for example people share links or try to access it by typing the URL or through the archive.

arpansahoo:
Sure, let me know if this works

static/js/play.js
```diff
@@ -49,6 +49,11 @@ let app = new Vue({
 
 
     mounted: async function() {
+        if (!("username" in serverData)) {
```
dqian3, line 52: This will prevent any logged in user from playing prompts, we only want this behavior for rated prompts. After looking at this though I realize that the redirect wasn't working due to a bug, so you don't have to worry about this anymore. I'll just revert it for you and then merge this.

PR #211, opened 2022-02-27
Play.js margin fix

mliu59:


dqian3 (changes requested):
There's a bit of a messiness here actually. Basically articleRenderer.js should be agnostic to what the timebox is. Either you should pass in the timebox element, or I think it would be better to call setMargin from the pageCallback.

Also, while you're at it would be good to do the same for the replay page from #216 which just got merged in

dqian3 (approved):

PR #213, opened 2022-02-27
User play info

dqian3:
Moved play link into separate col, and use played data to display play again/view results

image

### DIFF
--- a/apis/sprints.py
+++ b/apis/sprints.py
@@ -90,15 +90,20 @@ def get_managed_prompts():
 
 @sprint_api.get('/active')
 def get_active_prompts():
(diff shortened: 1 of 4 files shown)


dqian3:
Option 2:
image

mliu59:
Option 2 seems cleaner, at least from the screenshot. How do these look on mobile?

mliu59:
Suggest moving (play (again)) to the prompt # col
image

Will look better on mobile and also be consistent with where the link to play a prompt is

mliu59 (changes requested):
See comment above

dqian3:
image
image

PR #216, opened 2022-02-27
Allow users to replay paths from prompt leaderboard

codethejason:
Users are now able to click next to a now-formatted run path and go through the links in the path. The next step would be to scroll-to and highlight the following link in question, but I was not able to get to it. It seems like a difficult problem because we do not know which link corresponds to the topic we have (e.g. we do not know which link in an article corresponds with "Egg as food" without querying for more information).

https://user-images.githubusercontent.com/7005740/155879953-1352b915-dfe8-48d4-8b36-86a6cde953ec.mov


dqian3:
yo this is amazing

dqian3 (approved):

PR #217, opened 2022-02-27
Display link to leaderboard in archive page

arpansahoo:
I added a column in the archive table for links to leaderboard. Users can only see leaderboard links to games that they've already played. If they're not logged in, the column is just hidden altogether.

I also edited the get_archive_prompts function in prompts.py to left join the user's first run id for each prompt. With this, I set the leaderboard link to "/prompt/{prompt_id}?run_id={first_run}". I did this, as opposed to just simply checking whether they played the game, because I noticed the leaderboard points out your rank more clearly only when you have the run_id param in the web link

### DIFF
--- a/apis/sprints.py
+++ b/apis/sprints.py
@@ -98,7 +98,7 @@ def get_archive_prompts():
         limit = int(request.args.get('limit', 20))
         offset = int(request.args.get('offset', 0))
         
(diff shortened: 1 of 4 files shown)


dqian3 (commented):
Oh shoot, there's been a bit of miscommunication here unfortunately, I already did the equivalent with #213. I probably should have marked the issue.

It seems to me that this would have worked though oops.

PR #219, opened 2022-02-27
populate db script

dqian3:


mliu59:
Does this only generate data for rated prompts? Also would be helpful to add active, inactive (upcoming), and inactive (expired) prompts

bricehalder (approved):
this will be really useful, thanks

mliu59 (approved):

PR #225, opened 2022-02-28
edit readme to include /manage

codethejason:
Add line in README referring to `manage` so that developers are aware of the file when setting up prompts.


dqian3:
Great idea lol

PR #228, opened 2022-03-03
Leaderboard: If run_id isn't specified in web link, default to user's first run

arpansahoo:


wikispeedruns/prompts.py
```diff
@@ -129,8 +129,11 @@ def get_prompt(prompt_id: int, prompt_type: PromptType, user_id: Optional[int]=N
             return None
 
         if (user_id is not None):
-            cur.execute(f"SELECT COUNT(*) as count FROM {prompt_type}_runs WHERE prompt_id=%s AND user_id=%s", (prompt_id, user_id))
-            prompt["played"] = cur.fetchone()["count"] > 0
+            cur.execute(f"SELECT COUNT(*) as count, MIN(run_id) as first_run FROM {prompt_type}_runs WHERE prompt_id=%s AND user_id=%s",
```
bricehalder, line 132: I think `get_prompt_leaderboard` in `sprints.py` already does this work of computing the user's first run. Can we use that instead? I want to avoid having code in two different spots that does the same thing

bricehalder (changes requested):

wikispeedruns/prompts.py
arpansahoo: Good idea. I reverted my changes in `prompts.py`and just added the user's first run id to the response from `get_prompt_leaderboard`

bricehalder (approved):

Issue #214, opened 2022-02-27
Make different prompt leaderboards


dqian3:
1. Overall leaderboard (Current) - Display first run for everyone
2. Self (Default) - Best run(s) info + run history + graph of just own runs
3. Rated leaderboard (Default for rated) - Display first run if within rated window

PR #253, opened 2022-03-23
Anonymous sessions

mliu59:
Create localStorage of runs when not logged in. Push into db when logging in


dqian3 (changes requested):

static/js/play.js
```diff
@@ -9,13 +9,16 @@ these components should be as modular/generic as possible.
...
 import { fetchJson } from "./modules/fetch.js";
+import { startRun, submitRun } from "./modules/game/runs.js";
 
 import { CountdownTimer } from "./modules/game/countdown.js";
 import { FinishPage } from "./modules/game/finish.js";
 import { ArticleRenderer } from "./modules/game/articleRenderer.js";
 
 import { basicCannon, fireworks, side } from "./modules/confetti.js";
 
+import { addSprintRunToLocalStorage, submitSprintRunToLocalStorage, uploadAllLocalStorageSprintRuns } from "./modules/localStorage/localStorageSprint.js";
```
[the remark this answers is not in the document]
mliu59: renamed

mliu59:
> Quick question, do we use the locally stored runs at all? Besides just saving them and then uploading them later?

Not currently

mliu59:
Would want to tackle #291 before adding a personal leaderboard that includes all local runs, will be opening a separate PR for this

dqian3:
Haven't gotten a chance to look at the code thoroughly, but one concern I have is that local runs can't see their place in the leaderboard anymore? Before we would use their local `run_id` and display **You** in the leaderboard

mliu59:
https://user-images.githubusercontent.com/61032765/161212001-095c467c-8af4-4ab4-9e10-b67f275ea9c1.mp4

demo of main features. Marathon is identical to sprint

mliu59:
> Haven't gotten a chance to look at the code thoroughly, but one concern I have is that local runs can't see their place in the leaderboard anymore? Before we would use their local `run_id` and display **You** in the leaderboard

This is intentional. I did not want to work too much with tables before resolving #291 

So for now, anonymous runs will not show up at all on the leaderboard anywhere. 

In the future I plan on adding a separate personal leaderboard table that will include all anonymous runs / all personal runs on a prompt if logged in. #214

dqian3:
I still think we should have anonymous runs be able to be compared in the leaderboard. At the very least a user should be able to see see their run in the same page as the leaderboard. It might be worth moving the finishPage out of play.html, and have it be part of prompt.html if there is a `run_id` in the url. @maggarwal99 @bricehalder Any thoughts?

apis/marathon.py
```diff
@@ -44,50 +44,33 @@ def create_run():
...
-    nperbatch = int(request.json.get("nperbatch"))
-    
-    if nbucket < 1 or nbatch < 1 or nperbatch < 1:
-        raise ValueError("Bad input")
+    query = 'UPDATE `marathonruns` SET `user_id`=%s WHERE `run_id`=%s AND `user_id` IS NULL' 
     
+    if ('user_id' in session):
+        user_id = session['user_id']
+    else:
+        return f'Error updating marathon runs, user not logged in', 500
```
dqian3, line 59: `500` isn't the right error code here, we want to reserve this for unexpected errors

apis/marathon.py
```diff
@@ -44,50 +44,33 @@ def create_run():
...
-
-"""
-@marathon_api.post('/gen/')
-@check_admin
-def create_marathon_prompt():
-    
-    print("Received marathon prompt req")
-    print(request.json)
     
+@marathon_api.patch('/update_anonymous')
```
dqian3, line 48: Use the `@check_request_json` decorator in [`util/decorators.py`](util/decorators.py) to check the request json here.

apis/marathon.py
```diff
@@ -44,50 +44,33 @@ def create_run():
...
-    
-    initcheckpoints = [cp1, cp2, cp3, cp4, cp5]
-    
-    print("Starting prompt generation")
-    
-    checkpoints = genPrompts(initcheckpoints, batches=nbatch, nPerBatch=nperbatch, buckets=nbucket)
+    run_id = request.json['run_id']
     
-    print(checkpoints)
+    print(user_id, run_id)
```
dqian3, line 63: extraneous prints

apis/marathon.py
```diff
@@ -44,50 +44,33 @@ def create_run():
...
+
+    db = get_db()
+    with db.cursor() as cursor:
+        cursor.execute(query, (user_id, run_id))
+        db.commit()
+
+        return f'Updated marathon run {run_id} to user {user_id}', 200
+
+    return f'Error updating run {run_id} to user {user_id}', 500
```
dqian3, line 72: same thing about error code here

apis/runs.py
```diff
@@ -68,6 +68,33 @@ def finish_run(id):
     return f'Error updating run {id}', 500
 
 
+@run_api.patch('/update_anonymous')
```
dqian3, line 71: See the comments about `update_anonymous` for sprints

static/js/modules/localStorage/localStorageSprint.js
```diff
@@ -0,0 +1,59 @@
+import {startRun, submitRun, updateAnonymousRun} from "../game/runs.js"
+import { getLocalStorageRuns, addRunToLocalStorage, setLocalStorageRuns } from "./localStorage.js"
+
+function startLocalRun(promptId, runId) {
```
dqian3, line 4: So this is stored by runId? What do you think about storing by promptId first?

dqian3 (changes requested):

static/js/modules/localStorage/localStorageSprint.js
mliu59: I had it based on runId since its guaranteed to be unique, and it makes more sense as an iterable for when we pop all entries and make an API call for each run. Would there be a benefit for storing by promptId?

Also there may be more than one run per prompt

apis/marathon.py
mliu59: resolved

PR #256, opened 2022-03-23
added filter leaderboard by time interval (1 day, 7 days, 30 days, 100 days, or all (default))

mliu59:


apis/sprints.py
```diff
@@ -156,7 +156,7 @@ def get_prompt_leaderboard(id, run_id):
 
     # Then query the leaderboard (ew)
     query = '''
-    SELECT run_id, path, runs.user_id, username, TIMESTAMPDIFF(MICROSECOND, runs.start_time, runs.end_time) AS run_time
+    SELECT run_id, path, runs.user_id, username, TIMESTAMPDIFF(MICROSECOND, runs.start_time, runs.end_time) AS run_time, runs.end_time AS end_time
```
bricehalder, line 159: minor: any reason to explicitly use end time instead of the start time? I think it makes more sense to use the start time as the date you played the prompt

bricehalder: Also, could you upload a video or pictures to show how this currently looks like, would be useful for UI suggestions and stuff

mliu59: No particular reason why I chose end_time, we can switch to start_time instead. 

As for the demo, could use populate_db.py for displaying runs for more than 1 test user lmao :))))) @dqian3

dqian3: Yeah I think start time makes more sense, but not a big deal

static/js/prompt.js
```diff
@@ -317,14 +319,10 @@ var app = new Vue({
...
-            if (this.run_id) {
-                base += "&run_id=" + this.run_id
-            }
-            if (this.sortMode === 'path') {
-                base += "&sort=path"
-            }
-            window.location.replace(base)
+
+            let url = new URL(window.location.href)
+            url.searchParams.set('page', String(page))
```
dqian3, line 324: Ooh this is a lot cleaner

static/js/prompt.js
```diff
@@ -347,23 +345,50 @@ var app = new Vue({
...
+        filterByTime: function() {
+
+            let now = Date.now()
+
+            this.runs.forEach(el => {
+                let date = Date.parse(el.end_time)
+                console.log((now - date) / (1000 * 60 * 60 * 24))
+            });
+
+            if (!['1', '7', '30', '100'].includes(this.timeFilter)) return;
```
dqian3, line 379: Doing the filtering on the front end is fine here I think, but eventually we want to move it to the backend

dqian3 (approved):

PR #258, opened 2022-03-26
prevent loops from being added to path (only append to path if the new page wasn't the most recent visited page)

mliu59:
I believe this issue comes from a specific case when there is a link on a page that links to an anchored id on the same page, but does it through a different URL and then a redirect, so its being treated as if its an unique article (adding to path, counting clicks, etc.)

Not sure if there's a good way to handle extra anchor tags that are being treated as new pages (since the URLs might be different through a redirect).

We can instead prevent these loop edges from being counted towards things like the path or clicks for marathon

### DIFF
--- a/static/js/marathon.js
+++ b/static/js/marathon.js
@@ -130,10 +130,7 @@ let app = new Vue({
 
         async pageCallback(page, loadTime) {
 
-            if (this.path.length == 0) {
-                this.path.push(page);
-                this.clicksRemaining -= 1;
-            } else if (this.path[this.path.length - 1] != page) {
+            if (this.path.length == 0 || this.path[this.path.length - 1] != page) {
                 this.path.push(page);
                 this.clicksRemaining -= 1;
             }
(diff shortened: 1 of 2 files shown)


static/js/marathon.js
```diff
@@ -130,11 +130,15 @@ let app = new Vue({
 
         async pageCallback(page, loadTime) {
 
-            // console.log("page callback")
-
-            this.clicksRemaining -= 1;
+            if (this.path.length == 0) {
+                this.path.push(page);
+                this.clicksRemaining -= 1;
+            } else if (this.path[this.path.length - 1] != page) {
```
dqian3, line 136: Why not just use an or and short circuiting for this?

dqian3 (changes requested):

static/js/marathon.js
mliu59: forgot that was a thing lmao, fixed

dqian3 (approved):

PR #273, opened 2022-03-28
Add previews to marathon and refactor articleRenderer

maggarwal99:
1. Moved preview pages logic to articleRenderer.js so it can be shared between sprint and marathon mode
2. Refactored articleRenderer.js and play.js: mainly stuff concerning callbacks, async functions, and bindings

### DIFF
--- a/static/js/marathon.js
+++ b/static/js/marathon.js
@@ -1,6 +1,6 @@
 //JS module imports
 import { serverData } from "./modules/serverData.js";
-
+import { getArticleSummary } from "./modules/wikipediaAPI/util.js";
 import { submitRun, saveRun, loadRun, removeSave } from "./modules/game/marathon/runs.js";
 
 import { CountdownTimer } from "./modules/game/marathon/countdown.js";
@@ -34,6 +34,7 @@ let app = new Vue({
         startArticle: "",    //For all game modes, this is the first article to load
         lastArticle: "",
         currentArticle: "",
+        articlePreview: "",
         path: [],             //array to store the user's current path so far, submitted with run
 
         promptId: 0,        //Unique prompt id to load, this should be identical to 'const PROMPT_ID', but is mostly used for display
@@ -57,6 +58,11 @@ let app = new Vue({
         saved: false,
 
         renderer: null,
+        hover: false,
+        loading: false,
+
+        clientX: 0,
+        clientY: 0
     },
 
     computed: {
@@ -122,13 +128,16 @@ let app = new Vue({
         }
 
         this.startArticle = prompt['start'];
-        this.renderer = new ArticleRenderer(document.getElementById("wikipedia-frame"), this.pageCallback, this.setupPreview);
+        this.renderer = new ArticleRenderer(document.getElementById("wikipedia-frame"), this.pageCallback, this.mouseEnter, this.mouseLeave);
     },
 
 
     methods : {
 
-        async pageCallback(page, loadTime) {
+        pageCallback: function(page, loadTime) {
+
+            this.loading = false;
+            this.hover = false;
 
             if (this.path.length == 0 || this.path[this.path.length - 1] != page) {
                 this.path.push(page);
@@ -152,13 +161,6 @@ let app = new Vue({
                     this.visitedCheckpoints.push(page);
                 }
             }
-                       
-            if (this.clicksRemaining === 0) {
-                await this.finish(1);
-            }
-
-            setMargin();
-            
             
             if (hitcheckpoint) {
                 let el = this.checkpoints.shift()
@@ -166,11 +168,14 @@ let app = new Vue({
                 this.activeCheckpoints[checkpointindex] = el
 
                 conf();
-            }
 
-            if (!this.reachedstop && this.checkpointMarkReached) {
-                this.showStop = true
-                this.reachedstop = !this.reachedstop
+                if (!this.reachedstop && this.checkpointMarkReached) {
+                    this.showStop = true
+                    this.reachedstop = true
+                }
+
+            } else if (this.clicksRemaining === 0) {
+                this.finish(1);
             }
         },
 
@@ -186,10 +191,8 @@ let app = new Vue({
 
             if (load_save) {
                 await this.renderer.loadPage(this.lastArticle);
-                await this.pageCallback(this.lastArticle, Date.now() - this.startTime)
             } else {
                 await this.renderer.loadPage(this.startArticle);
-                await this.pageCallback(this.startArticle, Date.now() - this.startTime)
             }
 
             this.started = true;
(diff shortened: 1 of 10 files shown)


maggarwal99:
Temporarily removing sprint mobile UI changes. Will add them back for both sprint and marathon afterward.

maggarwal99:
Could you guys test out the changes as well since I refactored some stuff into articleRenderer so they can be reused between sprint and marathon? Tyty

dqian3:
Could you update this with a summary with an summary of the changes

static/js/marathon.js
```diff
@@ -152,25 +161,21 @@ let app = new Vue({
...
-            if (!this.reachedstop && this.checkpointMarkReached) {
-                this.showStop = true
-                this.reachedstop = !this.reachedstop
+                if (!this.reachedstop && this.checkpointMarkReached) {
+                    this.showStop = true
+                    this.reachedstop = true
+                }
+
+            } else if (this.clicksRemaining === 0) {
+                this.finish(1);
```
mliu59, line 178: Was there a specific reason to make pageCallback not async?

mliu59: Otherwise, agreed with moving finish to last, it makes more sense

static/js/modules/game/articleRenderer.js
```diff
@@ -5,47 +5,44 @@ export class ArticleRenderer {
...
+    constructor(frame, pageCallback, mouseEnter, mouseLeave) {
         this.frame = frame;
         this.pageCallback = pageCallback;
-        this.setupPreviews = setupPreviews;
+        this.mouseEnter = mouseEnter;
+        this.mouseLeave = mouseLeave;
     }
 
     async loadPage(page) {
+        try {
```
mliu59, line 16: I think many of the changes here is just a refactor to remove the extra loadPageWrapper we had before?

static/js/modules/game/articleRenderer.js
```diff
@@ -86,12 +81,16 @@ export class ArticleRenderer {
...
 
             // Remove "/wiki/" from string
-            this.loadPageWrapper(linkEl.getAttribute("href").substring(6))
+            this.loadPage(linkEl.getAttribute("href").substring(6))
         }
     }
 }
 
-
+function setMargin(frame) {
```
mliu59, line 89: Depending on how we choose to format mobile, the timebox element's ID might not always be "time-box"

mliu59 (commented):
Overall looks good. Tested locally and did not run into any issues with marathon or sprints. Still curious about some questions I left in the review

static/js/modules/game/articleRenderer.js
```diff
@@ -5,47 +5,44 @@ export class ArticleRenderer {
...
-            el.onclick = (e) => this.handleWikipediaLink(e);
-        });
 
-        window.scrollTo(0, 0);
+            hideElements(this.frame);
+            // disableFindableLinks(this.frame);
+            stripNamespaceLinks(this.frame);
 
-        return body["title"]
-    }
```
mliu59, line 39: I think many of the changes here is just a refactor to remove the extra loadPageWrapper we had before?

static/js/modules/game/articleRenderer.js
```diff
@@ -86,12 +81,16 @@ export class ArticleRenderer {
...
         }
     }
 }
 
-
+function setMargin(frame) {
+    const element = document.getElementById("time-box");
+    let margin = (element.offsetHeight + 25) > 250 ? (element.offsetHeight + 25) : 250;
+    frame.style.marginBottom = margin +"px";
+}
```
mliu59, line 93: Depending on how we choose to format mobile, the timebox element's ID might not always be "time-box"

mliu59 (commented):
some updates to line references

maggarwal99:
@dqian3 Updated the description.

static/js/modules/game/articleRenderer.js
maggarwal99: Yep yep

static/js/modules/game/articleRenderer.js
maggarwal99: Wouldn't the element ID stay the same? Or are you saying we might want to create/use a separate component for mobile?

static/js/marathon.js
maggarwal99: The only async logic inside pageCallback is that for the `saveRun` API call inside the `finish` method. Since this is a terminal non-blocking call it doesn't make sense to make pageCallback async otherwise we would need to `await` it everywhere we want to call it synchronously (which is important when rendering components that depend on the completion of the Vue tick cycle e.g. time-box margin). Moreover, since `await` can only be used inside async functions, we would be forced to make the calling function async as well then.

static/js/modules/game/articleRenderer.js
mliu59: Yeah, I wanted to potentially keep creating a separate component for mobile as an option, depending on how different it ends up being compared to the desktop one. 

But I do think that might be part of a greater discussion, and we can just deal with margin resize when the time comes lol

mliu59 (changes requested):
Can we bring back the mobile UI changes for sprint? Marathon's mobile UI is a bti more annoying to deal with and may also take more time, so we might want the sprint improvements implemented for the time being

maggarwal99:
> Can we bring back the mobile UI changes for sprint? Marathon's mobile UI is a bti more annoying to deal with and may also take more time, so we might want the sprint improvements implemented for the time being

Got it yep yep will revise this PR with mobile changes

static/js/modules/game/articleRenderer.js
maggarwal99: Ah kk, yea we can discuss that

static/js/marathon.js
```diff
@@ -34,6 +34,7 @@ let app = new Vue({
         startArticle: "",    //For all game modes, this is the first article to load
         lastArticle: "",
         currentArticle: "",
+        articlePreview: "",
```
dqian3, line 37: initializing it here as an empty string is rather confusing. I think either use `null` or expand out the expected fields like so
```
articlePreview: {
    image: null,
    html: null
}
```

static/js/marathon.js
```diff
@@ -57,6 +58,12 @@ let app = new Vue({
         saved: false,
 
         renderer: null,
+        hover: false,
+        loading: false,
+        blocker: false,
```
dqian3, line 63: what does this do?

static/js/modules/game/articleRenderer.js
```diff
@@ -86,12 +81,17 @@ export class ArticleRenderer {
...
     }
 }
 
-
+function setMargin(frame) {
+    const desktop_ht = document.getElementById("time-box") ? document.getElementById("time-box").offsetHeight : 225;
+    const mobile_ht = document.getElementById("time-box-mobile") ? document.getElementById("time-box-mobile").offsetHeight : 225;
+    const margin = Math.max(desktop_ht, mobile_ht, 225) + 25;
+    frame.style.marginBottom = margin + "px";
+}
```
dqian3, line 94: hmm, I don't like it relying on setting the margin by assuming these classes exist. Why not just do it in the callback (in the games)? Cause there's nothing that the article renderer has that is used to compute this margin?

static/js/play.js
```diff
@@ -140,21 +139,9 @@ let app = new Vue({
 
             this.startTime += loadTime;
 
-            setMargin();
-
             //if the page's title matches that of the end article, finish the game, and submit the run
             if (page.replace("_", " ").toLowerCase() === this.endArticle.replace("_", " ").toLowerCase()) {
```
dqian3, line 143: Do we still need this? Or can we just match directly?

static/js/marathon.js
maggarwal99: These are all states for previews

1. hover: determines whether or not to show the preview
2. loading: ensures that previews don't get stuck by handling edge cases like when the cursor is moving too fast
3. blocker: indicates whether the preview is for timebox or embedded hyperlinks since previews are triggered on hover for embedded hyperlinks and on click for the timebox. This creates a clickable overlay that allows a player to close the preview box by clicking anywhere outside it as opposed to having a dedicated close button

static/js/modules/game/articleRenderer.js
maggarwal99: I mean that's the point of the conditional:/ Can definitely do it in the callback but I want to limit marathon and play.js to the actual game logic as much as possible. Plus this way there's no code duplication. What's the problem with the conditional?

static/js/play.js
maggarwal99: Yep I think with the new API we can just match directly, will change

static/js/modules/game/articleRenderer.js
```diff
@@ -5,47 +5,44 @@ export class ArticleRenderer {
...
+            // disableFindableLinks(this.frame);
+            stripNamespaceLinks(this.frame);
 
-        return body["title"]
-    }
 
-    async loadPageWrapper(page) {
+            this.frame.querySelectorAll("a, area").forEach((el) => {
+                // Arrow function to prevent this from being overwritten
+                el.onclick = (e) => this.handleWikipediaLink(e);
```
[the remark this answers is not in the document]
maggarwal99: Yeah this part can def be refactored, it's a lil messy I can try to take a look but no commitment that I'll do it in this PR

static/js/marathon.js
maggarwal99: Good point, will use `null`

static/js/modules/game/articleRenderer.js
dqian3: I agree with hardcoding it actually, just put it to something pretty big lol

static/js/modules/game/pagePreview.js
```diff
@@ -0,0 +1,42 @@
...
+                styleObject['left'] = `${Math.floor((vw-360)/2)}px`;
+                styleObject['bottom'] = `${vh-this.clientY+25}px`;
+            }
+            return styleObject;
+        }
+	},
+
+	template: (`
+	<div class="tooltip-container" :style="computePosition()">
+        <img v-if="articlePreview.originalimage" :src="articlePreview.originalimage.source" />
```
mliu59, line 36: Instead of 'original-image', can we use 'thumbnail' instead? the original image is often a couple times larger than thumbnail, and takes considerably longer to load

See: https://en.wikipedia.org/api/rest_v1/page/summary/United_States as an exmaple

dqian3: I can make this change

dqian3 (approved):

PR #292, opened 2022-03-30
Add 0.6s delay to previews

maggarwal99:


dqian3 (approved):
I like the renaming of the component data for sure @maggarwal99, it's much clearer than before. I like the 0.6s too, it's not too long

PR #303, opened 2022-04-09
Warn players about their run being rated

dqian3:
HUD:
image
image

This field doesn't show up on lobby prompts though

Countdown:
image

### DIFF
--- a/static/js/modules/game/countdown.js
+++ b/static/js/modules/game/countdown.js
@@ -6,6 +6,7 @@ var CountdownTimer = {
     props: [
         "startArticle",
         "endArticle",
+        "ratedRun",
     ],
 
     data: function () {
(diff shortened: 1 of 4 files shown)


mliu59:
Does this show on private lobby prompts? 

A few things that I think could be cool to augment, but not necessary
- red text on HUD for "Rated" if rated
- extend countdown timer as we add more text to countdown page
- Some text on countdown page when not rated as well, just to clarify that this attempt will not be sent to the global leaderboard

dqian3:
> A few things that I think could be cool to augment, but not necessary, wanted to get your thoughts on these
> 
> * red text on HUD for "Rated" if rated
> * extend countdown timer as we add more text to countdown page
> * Some text on countdown page when not rated as well, just to clarify that this attempt will not be sent to the global leaderboard

I'll do the red text for rated, that's a good idea. I think extending the countdown is unnecessary, tbh I feel like most people skip anyways. As for the message, any idea of what to put? I was thinking "This is a practice attempt: your score will not appear on the public leaderboard"

mliu59 (approved):
I think overall we could do a better job explaining the rules (not specific to this PR), but I think that'd be a part of a greater effort to overhaul and improve out home page and UI in general.

PR #312, opened 2022-04-12
Fix wikipedia article mobile display

### LINKED ISSUE #311: Polish mobile wiki UI
Fixes to how wikipedia article itself displays on mobile

e.g.

image

maggarwal99:
Fixes how wikipedia articles display on mobile

Closes #311


dqian3 (approved):
yooooooo

dqian3:
This is exactly the kind of stuff that we're looking for, gonna merge into main to try it out on beta. 

As for the LaTeX stuff, it seems mobile wikipedia has scroll overflow. Is that something that we can fix easily @maggarwal99 ?

maggarwal99:
Ahh thanks for catching these guys! Can def fix the sizing and padding of images. Not sure about latex overflow tho, would need to look into it.

maggarwal99:
Hmm so looks like even the actual wikipedia mobile rendering of the polynomials page is messed up :/

The problem is that Wikipedia maintains different stylesheets for different use cases. And there's no single stylesheet that contains all of them (probably because of merge conflicts). So the only way to get the exact stylesheet for a given page is to parse its metadata from its html. This in itself isn't a problem but I strongly think that we shouldn't inject stylesheets in our html head mainly because that would force a complete rerender every time. Plus the stylesheets would need to be fetched from the web which would result in further unnecessary network load. This can easily add 1-2 s of latency to our game per page which frankly I don't think is worth the value it fetches in return (the value being a cleaner UI for edge cases). In any case I'll make some CSS changes to handle the latex overflow at least.

PR #317, opened 2022-04-13
Update play.html

mliu59:
Removed start article for both timeboxes, removed current article and links clicked on mobile


bricehalder (approved):
This will be a really good quality of life adjustment - thanks!

PR #319, opened 2022-04-13
added streaks for potd

mliu59:
Logged in users can now see a streak counter at the top of the home page. Flavored text will show up depending on:
- what the length of their streak is
- whether their last active potd play was today or yesterday

Users not logged in will not see this alert banner
image

The SQL query is a bit messy, so I will try to explain here:
1. get all potd active runs by user
2. group them by day of the potd and sort in ascending order, add a row number column
3. calculate the date difference between current date and the potd date as datediff
4. streak count is then the number of rows where datediff <= the row number


static/js/modules/streaks.js
```diff
@@ -0,0 +1,29 @@
...
+        return "You currently don't have an active streak. Play the Prompt of the Day to start your streak!";
+    }
+    let first_blurbs = {
+        5: `You currently have a 🔥<b> ${data['streak']} day streak </b>🔥, keep it up!`,
+        10: `Good to see you again! You currently have a 🔥<b> ${data['streak']} day streak </b>🔥, impressive!`,
+        25: `Woah, 🔥<b> ${data['streak']} day streak </b>🔥 and going strong!!`,
+    };
+
+    let disp = `Didn't think you'd make it this far, but you did! 🔥<b> ${data['streak']} day streak </b>🔥!!`
+    for (const threshold in first_blurbs) {
```
maggarwal99, line 16: for...in... loops don't maintain order when iterating over a hashtable in JS. Should use an array of objects or if...else... statements instead https://stackoverflow.com/questions/280713/elements-order-in-a-for-in-loop

maggarwal99 (approved):

static/js/modules/streaks.js
mliu59: Oh shit I had no idea, I will change that

dqian3 (approved):
LGTM

dqian3 (changes requested):
Actually I'm gonna add some tests

dqian3 (approved):

PR #322, opened 2022-04-16
Add total marathons played/finished and avg finished sprint runs per user

bricehalder:


mliu59:
Did we recently change our DB schema? Daily run api seems to be fetching an empty JSON object for my local instance

bricehalder:
> Did we recently change our DB schema? Daily run api seems to be fetching an empty JSON object for my local instance

Not that I know of, I have some DB changes incoming in a later PR though

dqian3 (approved):

PR #338, opened 2022-05-05
setup separate finish page

mliu59:
Separated out the post-game summary as a separate page, main changes:
- Finish page is now at `/finish` or `lobby/<lobby_id>/finish`. Run id and lobby id (when applicable) are passed as args
- there is also an optional arg `played` passed into the finish page. If false (default), some text such as "you found it" and "your run was submitted to the leaderboard", as well as the confetti fireworks will not show. This is for when we want to link players back to this specific summary page later well after they've finished this run
- Global leaderboard button is removed. Lobby finish page still has the button to go to the lobby leaderboard
- Lobby finish page also does not have the share buttons (can be changed), and the "return to home" button is changed to "back to lobby"
- Added an API endpoint for retrieving only one lobby run based on runid and lobbyid

Other changes also included in this PR:
- added a path share button (text can be beautified)
- removed some leftover console outputs in some pages

### DIFF
--- a/apis/lobbys.py
+++ b/apis/lobbys.py
@@ -157,3 +157,16 @@ def get_lobby_runs(lobby_id, prompt_id):
     runs = lobbys.get_lobby_runs(lobby_id, prompt_id)
 
     return jsonify(runs), 200
+
+
+# Run
+@lobby_api.get("/<int:lobby_id>/run/<int:run_id>")
+def get_lobby_run(lobby_id, run_id):
+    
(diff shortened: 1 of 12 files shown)


dqian3:
One small thing I want to comment on before I get into the actual code, but I feel like `/finish/` could be more descriptive. Why not keep `/prompt/`? And it could redirect to `/play/` when you haven't played it before.

dqian3 (approved):

PR #340, opened 2022-05-06
Finish tutorial

dqian3:


mliu59:
couple of small changes that I think might help, feel free to add/remove:
- Add a couple of introductory cards at the beginning (welcome, brief summary of rules and objectives of the game)
- Add another intro card specifically for the prompt
- Make active buttons more prominent (maybe use bootstrap's primary buttons)
- The first highlight doesn't seem to show 
- Ctrl+F wouldn't apply to mobile users, so we should rephrase that 
- For mobile, we should add an animation/icon to indicate swiping to advance instead of just text
- "Click the link to continue" should be more context specific ("Click on 'United States' to continue")

PR #348, opened 2022-05-11
Partial Runs 

### LINKED ISSUE #54: Modify path JSON to support partial run information

bricehalder:
### New JSON Format for Run Paths
```
    {
        "version": number
        "path": [
            ...
            {
                "article": string,
                "timeReached": number, 
                "loadTime": number,
            },
            ...
        ]
    }
```

We now store the db schema version, along with a JSON for tracking the system time of reaching a page as well as the time it took to load that page. This model is extensible so we could e.g. add support for time the page is in focus (in case someone tabs away and has the prompt going)

* Run path is updated on the frontend and async sent on each page visit, as well as when the tab is closed/goes out of view (good for tracking mobile)

* Added support + cleared up endpoint responsibilities (allows for differentiation for runs that are started but don't make it past countdown)
  * Upon starting a run simply creates a db entry
  * As pages are loaded, the entry is updated with start/end times and path

### DB Schema Changes
* `finished`: flag in `sprint_runs` and `lobby_runs`
* `play_time`: total prompt duration - cumulative sum of load times per page
   * Replaces the old `run_time` calculation we do which was essentially just total prompt duration
* `path`: Is still JSON, but the format has been changed to the above, no current enforcement of this schema
* A version number for the schema has been added. In order to retrieve this, `db.py` is made that should be kept in sync with the version in `schema.sql` (if there is an easier way to keep these in sync, please lmk) 
  * `CODEOWNERS` file added to notify me of any changes to the schema, to make sure that this is maintained

### Frontend Changes
* Support for parsing new path JSON in leaderboards

* Timer during a run now uses the current duration of the prompt - cumulative sum of load times, this should be more accurate to the final result of `play_time`. `start_time` is no longer incremented since that was a hacky workaround.

### Testing
TBD will add video later but works on my machine™

## Merging to Prod
Might want to do this together (@dqian3)
```
ALTER TABLE sprint_runs ADD finished BOOL DEFAULT FALSE AFTER end_time;

ALTER TABLE lobby_runs ADD finished BOOL DEFAULT FALSE AFTER end_time;



UPDATE sprint_runs SET finished=TRUE WHERE end_time IS NOT NULL;

UPDATE lobby_runs SET finished=TRUE WHERE end_time IS NOT NULL;



ALTER TABLE sprint_runs ADD play_time FLOAT AFTER end_time;

ALTER TABLE lobby_runs ADD play_time FLOAT AFTER end_time;



UPDATE sprint_runs SET play_time=TIMESTAMPDIFF(MICROSECOND, start_time, end_time) / 1000000  WHERE play_time IS NULL AND finished IS TRUE;

UPDATE lobby_runs SET play_time=TIMESTAMPDIFF(MICROSECOND, start_time, end_time) / 1000000  WHERE play_time IS NULL AND finished IS TRUE;

ALTER TABLE lobby_runs MODIFY `start_time` TIMESTAMP(3) NULL;
ALTER TABLE lobby_runs MODIFY `end_time` TIMESTAMP(3) NULL;
ALTER TABLE lobby_runs MODIFY `path` JSON NULL;
```

Closes #54


mliu59:
Are changes to locally stored runs (anonymous users) included? Or would any new run be saved into localStorage with this new schema?

bricehalder:
> Are changes to locally stored runs (anonymous users) included? Or would any new run be saved into localStorage with this new schema?

From inspection it looks like the `localStorage` and `localStorageSprint.js` simply store and get the path JSON (but not actually interact with the JSON internals), so it seems like it should work and be serialized/deserialized as expected.

apis/sprints.py
```diff
@@ -193,7 +192,9 @@ def get_prompt_leaderboard(id, run_id):
         results = cursor.fetchall()
 
         for run in results:
-            run['path'] = json.loads(run['path'])
+            pathJson = json.loads(run['path'])['path']
+            run['path'] = [entry['article'] for entry in pathJson]  
+            print(run['path'])
```
dqian3, line 197: extraneous print

scripts/schema.sql
```diff
@@ -30,7 +39,23 @@ CREATE TABLE IF NOT EXISTS `sprint_runs` (
     `run_id` INT NOT NULL AUTO_INCREMENT,
     `start_time` TIMESTAMP(3) NULL,
     `end_time` TIMESTAMP(3) NULL,
+    `play_time` FLOAT,
```
dqian3, line 42: Could this be a computed property instead? Or does it also take into account the load times

dqian3 (changes requested):
Had a few small comments.

However the big thing is I was wondering if we should make the lobby run and the sprint run api/interaction more consistent. Maybe we can just combine the 2 and make lobby id an optional parameter. For example, right now lobby has add run, but that doesn't get updated while the user is playing it seems? Or actually it looks like it might post multiple runs everytime submitRun is called on the frontend?

Also still waiting on the script to change the old schemas.

static/js/play.js
```diff
@@ -114,13 +141,26 @@ let app = new Vue({
...
+                // Update path
+                let timeElapsed = (Date.now() - this.startTime) / 1000;
+                this.path.push({
+                    "article": page,
+                    "timeReached": timeElapsed,
+                    "loadTime": loadTimeSeconds
+                });
+
+                // Update partial run information for non-lobby runs
+                if (this.lobbyId === null) {
```
dqian3, line 156: Actually my point about potentially submitting multiple runs isn't right, but we should also update for lobby runs I think

scripts/schema.sql
bricehalder: This takes into account the load times! 

https://github.com/wikispeedruns/wikipedia-speedruns/pull/348/files#diff-f143ca6d25ffa21c96acaa8486f99b2a0c02bbd3d9e89c5c89195e678ac2b27eR164-R166

dqian3:
Another note, I feel like start_time should be set on `create_run`

bricehalder:
> Another note, I feel like start_time should be set on `create_run`

The reason start_time is not set in `create_run` is that the run is created when the user initially hits play/starts a run, whereas the start_time starts from the start of the actual play i.e. after the countdown. The start_time for a run is set when the starting article is initially loaded

dqian3:
I feel like we should still send the start time, especially if play time is
now separate and computed. Then start time can just be used to figure out
when the run takes place.

bricehalder:
> I feel like we should still send the start time

Yeah, I think it's a good idea, can include it in this PR but can also come in the future I think

Basically need to modify the path JSON to look like 

```
   {
        "version": number,
        "countdownTime": number,
        "path": [
            ...
            {
                "article": string,
                "timeReached": number, 
                "loadTime": number,
            },
            ...
        ]
    }
```

Right now `play_time` is computed as `finish_time - start_time - {cumulative load times}`
and with this new change it can be `finish_time - start_time - {cumulative load times} - countdownTime`

dqian3:
Note: I modified the migration script to include 

```
ALTER TABLE lobby_runs MODIFY `start_time` TIMESTAMP(3) NULL;
ALTER TABLE lobby_runs MODIFY `end_time` TIMESTAMP(3) NULL;
```

dqian3 (approved):

PR #356, opened 2022-05-24
hide lobby link/passcode by default

dqian3:
It's a little ugly, but lobby page needs an overhaul anyways

image

image


mliu59 (approved):
looks good

PR #357, opened 2022-05-24
Generator page and lobbies generator page

dqian3:


bricehalder:
I think this will be a good base to transition to personal leaderboards. i.e. replacing leaderboards with only your runs, but not necessarily on a separate page for now

bricehalder (changes requested):
See comment above - basically just change the text. Also, maybe it should be expanded by default?

mliu59:
Are there additional setup steps for local? I haven't been able to try it
- Agreed with all Brice has mentioned, especially "Obscurity" vs. "reachability"
- I still think we should increase the article pool size
- I think separate start/end should be the default, and the generator block should be a vue component/js module since we're gonna use it in multiple places

dqian3:
image

bricehalder:
Hmm, I think the numbers still make it confusing. e.g. what does 50 obscurity mean, or 3000 articles? It's not really obvious to a user I think. What about if we just had the sliding scale and had it look like:

More Common ----O------------- Less Common

or

Easy ---O------------- Hard

but I prefer the first one.

PR #362, opened 2022-05-28
Sprint hotfixes

mliu59:
1. Fixed bug that caused bad request error when submitting the first run update (upon loading first page). The issue was that when we preload the article prior to the countdown, startTime was not initialized. Solution: moved startTime to be initialized before the first article is loaded, and store the duration from this startTime to when the countdown finishes (automatically or manually) as the 'timeReached' for the first article. This value is then subtracted from the final 'play_time'. This also fixed the bug that caused the first path element to have seconds since epoch timeReached values. 
2. anonymous users weren't able to view finished run results due to unauthorized error. We probably want to keep the restrictive access to the finish page, so solution was to instead read from locally stored runs if user is not logged in on the finish page (see user reported screenshot on discord). Locally stored sprint runs are now also stored with the 'play_time' value, computed at the end of a run.

### DIFF
--- a/static/js/modules/localStorage/localStorageSprint.js
+++ b/static/js/modules/localStorage/localStorageSprint.js
@@ -21,6 +21,12 @@ function submitLocalRun(promptId, runId, startTime, endTime, finished, path) {
         path: path
     };
 
+    let totalloadtime = -path[0]['loadTime'] + path[0]['timeReached'];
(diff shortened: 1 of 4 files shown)


bricehalder (approved):
I think we'd want to change this to be a countdown_time column in the runs table but this works for now

PR #369, opened 2022-06-05
Fix archive + sort by prompt

### LINKED ISSUE #368: archive ordering is messed up
Archive is really messed up rn

bricehalder:
* Changes archive sorting to sort on frontend instead of doing another api call
  * Default sort is by release date
* Adds ability to sort by prompt number
* Removes reference to rated prompts

Closes #368 

Test:

https://user-images.githubusercontent.com/17424008/172035577-728e0dba-c718-429c-8599-a8affcf43326.mp4

### DIFF
--- a/apis/sprints.py
+++ b/apis/sprints.py
@@ -99,11 +99,9 @@ def get_archive_prompts():
     try:
         limit = int(request.args.get('limit', 20))
         offset = int(request.args.get('offset', 0))
-        sort_desc = request.args.get('sort_desc', "True").lower() == "true"
         sprints, num_prompts = prompts.get_archive_prompts("sprint",
             offset=offset,
             limit=limit,
-            sort_desc=sort_desc,
             user_id=session.get("user_id")
         )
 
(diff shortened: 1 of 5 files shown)


mliu59:
Is the pagination determined by release time? or prompt ID

bricehalder:
Pagination is by release date

dqian3 (approved):

dqian3:
Yeah actually before I merge, just to make sure I have this right, the filtering + sorting is  on the backend, but then the pagination is on the front?

bricehalder:
Other way around I think. Backend receives limit + offset api request
(pagination), which is always sorted by release date. The frontend can
change the sort for the prompts currently displayed between release date
and prompt number.

bricehalder:
It's also worth noting that I always treat sort_desc as true, since I couldn't figure out the underlying issue of its state not being persisted. 

But I think it makes sense this way (and hopefully will lead to less issues too since it's consistent)

PR #370, opened 2022-06-05
Lobby prompt deletion

### LINKED ISSUE #353: Allow lobby hosts to delete prompts

bricehalder:
Adds an edit mode for lobby owners
* Shows/hides checkboxes next to prompts
   * Can be used for other things in the future wrt prompt management
* Delete button to delete selected prompts
   * Pop-up alert added for confirmation
   * Backend checks that user session is owner of lobby

Deletion nukes the prompts and their runs from our database
Closes #353 

Test:

https://user-images.githubusercontent.com/17424008/172041671-7321fb79-d15c-4ea4-8851-c2d8470594f2.mov

View from non-owner (no edit button is shown):
Screen Shot 2022-06-05 at 1 08 46 AM

### DIFF
--- a/templates/lobbys/lobby.html
+++ b/templates/lobbys/lobby.html
@@ -133,6 +133,10 @@ var app = new Vue({
         },
 
         async deletePrompts() {
+            if (!confirm("Are you sure you want to delete the selected prompts and their runs?")) {
+               return; 
+            }
+
             const resp = await fetchJson(`/api/lobbys/${LOBBY_ID}/prompts`, "DELETE", {
                 "prompts": Array.from(this.selectedPrompts),
             }); 
@@ -239,7 +243,7 @@ var app = new Vue({
                     </tr>
                 </tbody>
             </table>
-            <input v-if="editPrompts" @click="deletePrompts" class="btn btn-danger" value="Delete Selected">
+            <input v-if="editPrompts" href="javascript:;" @click="deletePrompts" class="btn btn-danger" value="Delete Selected">
         </div>
     </div>
 
--- a/templates/lobbys/lobby.html
+++ b/templates/lobbys/lobby.html
@@ -11,6 +11,10 @@
 
 }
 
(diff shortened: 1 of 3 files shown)


templates/lobbys/lobby.html
```diff
@@ -199,9 +237,13 @@ <h3> Prompts </h3>
                         <td>[[prompt.end]]</td>
 
                         <td><a v-bind:href="'/lobby/' + lobbyInfo.lobby_id + '/prompt/' + prompt.prompt_id">Results</a></td>
+                        <td v-if="editPrompts">
+                            <input type="checkbox" @click="selectPrompt(prompt.prompt_id)">
+                        </td>
                     </tr>
                 </tbody>
             </table>
+            <input v-if="editPrompts" href="javascript:;" @click="deletePrompts" class="btn btn-danger" value="Delete Selected">
```
dqian3, line 246: why is this an input instead of a normal button?

templates/lobbys/lobby.html
```diff
@@ -199,9 +237,13 @@ <h3> Prompts </h3>
                         <td>[[prompt.end]]</td>
 
                         <td><a v-bind:href="'/lobby/' + lobbyInfo.lobby_id + '/prompt/' + prompt.prompt_id">Results</a></td>
+                        <td v-if="editPrompts">
+                            <input type="checkbox" @click="selectPrompt(prompt.prompt_id)">
```
dqian3, line 241: Rather than using on-click to manually maintain this state, v-model might be a bit cleaner

dqian3 (changes requested):

templates/lobbys/lobby.html
bricehalder: tried it as a normal button, looked too HTML-y, making it input let the bootstrap styles apply. Modeled it after the Submit button on the same page

dqian3 (approved):

PR #371, opened 2022-06-05
Pause timer on page load

bricehalder:
Purely frontend changes, pause/start a timer instead of subtracting the load time (which confusingly incremented during page load)
* Lines up similarly with backend times (not that it lined up exactly before)

Test:

https://user-images.githubusercontent.com/17424008/172046840-b502e2d0-180c-4b79-a589-e30d873612a8.mp4


dqian3:
nice

dqian3 (approved):

PR #374, opened 2022-06-08
Fix race condition and incorrect timebox time

bricehalder:
* Race condition between start() and first page callback
* Zero out timer on starting play


mliu59 (approved):
Ran a couple of prelim tests on local, everything seems good. Thanks for resolving the merge conflict!

PR #379, opened 2022-06-11
Account + profile page

mliu59:


mliu59:
I plan on adding some auto tests for the user deletion stuff

apis/users.py
```diff
@@ -330,6 +363,27 @@ def confirm_email():
     return "Email Confirmed"
 
 
+@user_api.post("/check_email_confirmation")
+def check_email_confirmation():
+    username = request.json
+
+    # user must be logged in to access
+    if session["username"] != username:
+        return "false"
```
dqian3, line 372: Give more meaningful responses here and below (i.e. inform the user the email is already confirmed), also make use of HTTP status codes (i.e. 400 when not logged in).

apis/users.py
```diff
@@ -291,6 +292,38 @@ def change_password():
...
+        try:
+            cursor.execute(update_query, (new_username, id))
+            db.commit()
+        
+        except pymysql.IntegrityError:
+            print("dup username")
+            return (f'Username `{new_username}` already exists', 409)
+        except pymysql.Error as e:
+            return ("Unknown error", 500)
+
```
dqian3, line 322: You should probably update the session here, otherwise the header won't update I think

apis/profiles.py
```diff
@@ -50,6 +50,7 @@ def get_total_stats(username):
     SELECT
         users.user_id,
         COUNT(run_id) AS total_runs,
+        COUNT(case finished when 1 then 1 else null end) AS total_completed_runs,
```
dqian3, line 53: dang is this really the sql case syntax?

apis/users.py
```diff
@@ -393,3 +447,91 @@ def reset_password():
...
+    args = {"user_id": id}
+
+    db = get_db()
+    with db.cursor(cursor=DictCursor) as cursor:
+        
+        count = cursor.execute(get_hosted_lobbies_query, args)
+        
+        lobbies = cursor.fetchall()
+        
+        if (count > 0):
```
dqian3, line 499: Rather than doing `... WHERE user_lobbys.lobby_id IN ...` you should probably use `cursor.executemany`

static/js/nav.js
```diff
@@ -24,6 +24,10 @@ var app = new Vue({
             event.preventDefault();
             await fetch("/api/users/logout", {method : "POST"});
             window.location.href = "/";
```
dqian3, line 26: If logout isn't used anymore, get rid of it here

static/js/modules/profileStats.js
```diff
@@ -0,0 +1,90 @@
+import { serverData } from "./serverData.js";
+
+var profileStatsTable = {
+
```
dqian3, line 4: messed up indents

static/js/modules/profileStats.js
```diff
@@ -0,0 +1,90 @@
...
+	data: function () {
+        return {
+            username:"",
+            basic_stats: {
```
dqian3, line 8: inconsistency with variable/field names, I think we usually use camelCase for JS

dqian3 (changes requested):

apis/profiles.py
mliu59: I couldnt figure out anything else for this, but theres probably a more elegant way that I'm not aware of. This works fine tho

apis/users.py
mliu59: Added

static/js/modules/profileStats.js
mliu59: Fixed

static/js/modules/profileStats.js
mliu59: Are you referring to the template html? if so, fixed

static/js/nav.js
mliu59: fixed

apis/users.py
mliu59: Fixed

apis/users.py
mliu59: Fixed

apis/profiles.py
mliu59: Here's where I got it: https://stackoverflow.com/questions/1400078/is-it-possible-to-specify-condition-in-count

mliu59:
@danielliu04 one issue with the current profile page (not account), is that achievements will always show the logged in user's achievements, instead of the profile owner,

Can you make the achievement table more flexible in that regard? Note that we access profiles with username and not user_id

dqian3 (approved):

PR #394, opened 2022-07-06
added lobby player list and change host for hosts

mliu59:
Added 3 endpoints for lobby API (only accessible by host):
- get all users in a lobby
- get all anon_users in a lobby
- change host to another user in the lobby

Added corresponding UI section for lobby hosts:
https://vimeo.com/727473707/31e5137af3

### DIFF
--- a/templates/lobbys/lobby.html
+++ b/templates/lobbys/lobby.html
@@ -19,6 +19,10 @@
     display: none;
     margin-left: 10px;
 }
+
+.active-dropdown {
+    display: block !important;
+}
 </style>
 
 
@@ -161,11 +165,18 @@ var app = new Vue({
 
 
         async getLobbyUsers() {
-            const resp = await fetchJson(`/api/lobbys/players/${LOBBY_ID}`, "GET");
-            this.users = await resp.json();
 
-            const resp2 = await fetchJson(`/api/lobbys/anon_players/${LOBBY_ID}`, "GET");
-            this.anon_users = await resp2.json();
+            Promise.all([
+                fetchJson(`/api/lobbys/players/${LOBBY_ID}`, "GET"), 
+                fetchJson(`/api/lobbys/anon_players/${LOBBY_ID}`, "GET")
+            ])
+            .then(responses =>
+                Promise.all(responses.map(res => res.json()))
+            ).then(resp => {
+                this.users = resp[0];
+                this.anon_users = resp[1];
+            }).catch(err => console.log(err));
+
         },
 
         async makeHost(user_id, username) {
@@ -184,6 +195,10 @@ var app = new Vue({
                 console.log(e)
                 alert(e);
             }
+        },
+
+        toggleDropdown(id) {
+            document.getElementById(id).classList.toggle("active-dropdown")
         }
     }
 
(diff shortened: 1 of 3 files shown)


templates/lobbys/lobby.html
```diff
@@ -177,10 +211,45 @@ <h2 v-else>Lobby [[lobbyInfo.lobby_id]]: </h2>
...
+        <h3>Lobby settings</h4>
+
+        <details>
+            <summary>List of players</summary>
+            <div class="card">
+                <div class="card-body">
+                    <table class="table table-hover">
+                        <thead>
+                            <th scope="col">Display name</th>
+                            <th scope="col">Host</th>
```
bricehalder, line 229: Is the host column necessary? I can see this being useful if we want to support multiple hosts in the future, but right now it seems like we just support one, in which case I think the column is confusing to have in the UI (since only the host sees this anyways).

Related - I think "Make host" is not clear right now, since with the Host column, it feels like you are _adding_ a new host rather than _transferring_ ownership. I think we should either 
1. change the functionality here to add an additional host
  1a. Change the backend/db schema to support a set of userid's rather than a single userid
2. change the text to reflect the transfer i.e. "Make host" --> "Transfer Ownership" or similar and make that clear in the pop-up as well.

templates/lobbys/lobby.html
```diff
@@ -177,10 +211,45 @@ <h2 v-else>Lobby [[lobbyInfo.lobby_id]]: </h2>
...
+                            <th scope="col">Full user</th>
+                            <th scope="col"></th>
+                        </thead>
+                        <tbody>
+                            <tr v-for="user in users" v-cloak>
+                                <td>[[user.username]]</td>
+                                <td><i v-if="user.owner" class="bi bi-check"></i></td>
+                                <td><i class="bi bi-check"></i></td>
+                                <td><button v-if="!user.owner" class="btn btn-link btn-sm" @click="makeHost(user.user_id, user.username)"><small>Make host</small></button></td>
```
bricehalder, line 238: See above comment regarding "Make host" text

But in addition:

(Personal preference) 
I think this should probably be a dropdown menu with options per user (for now, this can be the only thing in the dropdown).

1. It would be cleaner UI
2. We probably want to support more user management options in the future, namely kicking or banning users from the lobby, and this would be a clean and extensible way to support those additional options per user.

bricehalder (changes requested):

bricehalder:
Nice, I think the functional part looks good. Left some comments mostly around clarity of our UI

templates/lobbys/lobby.html
```diff
@@ -177,10 +211,45 @@ <h2 v-else>Lobby [[lobbyInfo.lobby_id]]: </h2>
...
+
+        <details>
+            <summary>List of players</summary>
+            <div class="card">
+                <div class="card-body">
+                    <table class="table table-hover">
+                        <thead>
+                            <th scope="col">Display name</th>
+                            <th scope="col">Host</th>
+                            <th scope="col">Full user</th>
```
bricehalder, line 230: (personal preference)

I think "Full user" is kind of confusing too. I think "Registered User" or "Anonymous User" (and flipping the condition) would be more clear

templates/lobbys/lobby.html
```diff
@@ -150,6 +157,33 @@
             this.editPrompts = false;
             this.clearSelectedPrompts();
             this.getPrompts();
+        },
+
+
+        async getLobbyUsers() {
+            const resp = await fetchJson(`/api/lobbys/players/${LOBBY_ID}`, "GET");
```
dqian3, line 164: Small thing, but these can be done in parallel using Promise.all().

mliu59:
Addressed comments, see image
image

The dropdown implementation is fairly crude, but bootstrap was giving me alot of issues with aria attributes, so I opted for a workaround

templates/lobbys/lobby.html
```diff
@@ -177,10 +226,58 @@ <h2 v-else>Lobby [[lobbyInfo.lobby_id]]: </h2>
...
+                                                data-toggle="dropdown" 
+                                                aria-haspopup="true" 
+                                                aria-expanded="true"
+                                                @click="toggleDropdown(String(user.user_id)+'dropdown')">
+                                            <small>Options</small>
+                                        </button>
+                                        <div class="dropdown-menu" 
+                                            v-bind:id="String(user.user_id)+'dropdown'">
+                                            <button v-if="!user.owner" class="btn btn-sm" @click="makeHost(user.user_id, user.username)">Transfer Ownership</button>
+                                            <button class="btn btn-sm disabled" v-else>Currently host</button>
```
bricehalder, line 264: (personal preference)
Don't really think the "Currently host" is needed since it's implied, so I would probably just delete this line. I was kind of confused looking at the example image.

bricehalder (approved):
Looks good, just another small comment

PR #398, opened 2022-07-08
Achievements

danielliu04:
These are all the updates for Achievements. The current achievements implemented can be seen in the Achievement List Google Docs (implemented and unimplemented achievements are separated with a line break)

To make sure the Achievements function as intended, here are some steps:

1. Run scripts to set up everything:
- create_db.py: Make sure that the two new tables (list_of_achievements, achievements_progress) are added to the database
- expand_sprint_runs.py: Expands sprint_runs table to add in new column `counted_for_am`, which is used to make sure the same run doesn't get counted multiple times towards achievements
- add_achievements_to_database.py: adds all achievements to list_of_achievements database

2. Optional:
- historical_achievements.py: Run this script to consider all past runs and update achievements accordingly (only run after previous scripts are finished)

Now achievements should be working as intended!

Functionalities:

1. Profile page:
As long as user is logged in (doesn't matter if it is admin or normal user), they would be able to see their achievements_progress in their profile page.
As of right now, the achievements are sorted in this order:
a. unlocked achievements (most recent achievement at the top)
b. locked achievements that are not hidden: in arbitrary order
c. locked achievements that are hidden

2. New run:
Currently, achievements are only considered for the daily prompt and the other prompts posted publicly for sprint_runs. Finishing a run will pull up new play_finish page that includes achievements progress at the bottom.


wikispeedruns/achievement_functions.py
```diff
@@ -0,0 +1,288 @@
...
+    return "Volkswagen emissions scandal" in single_run_article_map, None, None
+
+def taking_over_the_internet(single_run_data: Dict[str, Any], single_run_article_map: Dict[str, int], current_progress: Any) -> ReturnType:
+    return "Love Nwantiti" in single_run_article_map, None, None
+
+def you_lost(single_run_data: Dict[str, Any], single_run_article_map: Dict[str, int], current_progress: Any) -> ReturnType:
+    return sum(i >= 2 for i in list(single_run_article_map.values())) > 0, None, None
+
+def fastest_gun_alive(single_run_data: Dict[str, Any], single_run_article_map: Dict[str, int], current_progress: Any) -> ReturnType:
+    num_links = len(single_run_data["path"]) - 1
```
mliu59, line 98: This line gave me a divide by 0 error. Path could potentially be len 1 if the user immediately quits after the game starts so we should account for that

mliu59 (changes requested):
Overall looks great! Will be a great feature addition. 

Couple of general comments:
1. historical achievements script didn't work for me at first because I had a few runs in the sprint_runs table that were completely null except for run_id. I think this is just because my db was outdated, but we may want to make sure it doesn't happen for prod. We should add a simple format checker to catch and discard bad/outdated data when checking for achievements
2. Remove all test_data variables in the main code. 
3. Related to 2, we should also write a couple automated tests for this whole thing. See `test/` folder.
4. Instead of having the HTML run an additional js file, I think the logic used in `achievement_data.js` should be merged with the `achievement_table` component. I don't think there's any potential way we could use these two separately so they should be together

apis/achievements.py
```diff
@@ -0,0 +1,44 @@
...
+from db import get_db
+from pymysql.cursors import DictCursor
+
+from wikispeedruns import achievements
+
+achievements_api = Blueprint('achievements', __name__, url_prefix='/api/achievements')
+
+
+@achievements_api.get('/get/<int:run_id>')
+def get_new(run_id):
```
dqian3, line 14: I would recommend a more descriptive name here for the endpoint. `get` is a bit generic, also idk if it should be a get because the response would change after multiple runs. Maybe it should be a PATCH with a name related to `process`

apis/achievements.py
```diff
@@ -0,0 +1,44 @@
...
+
+        db.commit()
+
+    return new_achievements
+
+
+
+@achievements_api.get('/user')
+@check_user
+def get_all_achievements():
```
dqian3, line 36: Here. is an example of what a get should be more like.

scripts/expand_sprint_runs.py
```diff
@@ -0,0 +1,42 @@
+import pymysql
```
dqian3, line 1: In the future, these kinds of changes can just be implemented as a SQL script

scripts/remove_achievement.py
```diff
@@ -0,0 +1,67 @@
+import pymysql
```
dqian3, line 1: Can you put these script files in a subfolder perhaps?

static/achievements/metadata.js
```diff
@@ -0,0 +1,135 @@
+var achievement_list = {
```
dqian3, line 1: This is where the image paths would go I assume too?

dqian3 (commented):
Looks mostly good to me! Great work

I had a few very small things, but nothing major

dqian3:
> Overall looks great! Will be a great feature addition.
> 
> Couple of general comments:
> 
> 1. historical achievements script didn't work for me at first because I had a few runs in the sprint_runs table that were completely null except for run_id. I think this is just because my db was outdated, but we may want to make sure it doesn't happen for prod. We should add a simple format checker to catch and discard bad/outdated data when checking for achievements
> 2. Remove all test_data variables in the main code.
> 3. Related to 2, we should also write a couple automated tests for this whole thing. See `test/` folder.
> 4. Instead of having the HTML run an additional js file, I think the logic used in `achievement_data.js` should be merged with the `achievement_table` component. I don't think there's any potential way we could use these two separately so they should be together

For (2), agreed about making these into tests.

For (4) I actually disagree, I like having it be a separate data file. In fact we could argue that it shouldn't even be a js file, but I think it's convenient like this.

danielliu04:
Currently struggling with pytest, I think the main problem is that for achievements to work I have to also get a sprint_prompt and sprint_run down.

Made it so that there is a preliminary check for the run_data before being processed.

I have the processing of new achievements as "get" because it has to return new achievements. So it is essentially a patch and then a get. Putting get makes it convenient but I can try to implement something to change that.

dqian3:
> I have the processing of new achievements as "get" because it has to return new achievements. So it is essentially a patch and then a get. Putting get makes it convenient but I can try to implement something to change that.

You can still return something from a patch though?

PR #403, opened 2022-07-14
Achievements

danielliu04:
Overall there were a few changes made:

- Fixed the around the world achievement
- Fixed the achievement time problem (at least I think)
- Added a header for achievements in profile page (including an achieved/total)
- Added text "Hidden" for completed hidden achievements when displayed

We do need to rerun historical achievements to make sure the around the world achievement is accounted for.

I am not exactly sure why this can't be automatically merged.


danielliu04:
I do not think there were other changes made to these files, so they can all be replaced.

wikispeedruns/achievement_functions.py
```diff
@@ -145,10 +145,13 @@ def around_the_world_in_80_seconds(single_run_data: Dict[str, Any], single_run_a
             break
 
         article = path[i]["article"]
-        current_map[article] -= 1
-        if current_map[article] == 0:
-            del current_map[article]
-    
+        i += 1
+
+        if article in article_set:
```
dqian3, line 150: I didn't take a close look at this before, but this definitely seems a bit complicated. Could you add something in explaining it? From my impression, you're trying to find the subarray within the path that contains all 7 articles, but with minimum time? `i` is like a left index and `j` a right index?

dqian3 (approved):

PR #413, opened 2022-07-26
Setup scraper task queue

dqian3:
Also refactored some of the scraper stuff and internal modules


bricehalder (approved):
I don't really have much input/feedback but just looked this over and stamping to unblock

PR #416, opened 2022-08-01
Quick play + Autocomplete inputs

danielliu04:
This PR includes the introduction of a quick-play option with autoscroll as well as a new input element that autocompletes wikipedia articles.

Currently, quick-play functions almost identical to sprints, except it occupies different tables and local data, does not count for achievements, and requires a starting and ending prompt instead of a prompt_id. 

The autocomplete input uses data from wikipedia api to give suggestions for articles when searching. It is currently in static/js/modules and can be reused elsewhere. There is a bit of syntax to deal with, an example of it shown in quick_play.html.

Setup:
1. Create new database 'quick_runs` at the end of 'schema.sql' in scripts

Issues / Possible Changes:
2. Change the difficulty for quick play (maybe?). Currently its 1000, but it can be changed to whatever


apis/runs_api.py
```diff
@@ -30,17 +32,30 @@ def create_lobby_run(prompt_id, lobby_id):
     )
     return jsonify({"run_id": run_id})
 
+@run_api.post("/quick_runs/<string:prompt_start>/<string:prompt_end>/runs")
+def create_quick_run(prompt_start, prompt_end):
+    run_id = runs.create_quick_run(unquote(prompt_start), unquote(prompt_end), session.get("user_id"))
```
dqian3, line 37: what's the need for the unquote thing here? Are there potentially quotes within the title?

views.py
```diff
@@ -130,7 +142,7 @@ def get_sprint_finish_page():
     try:
         run_id = int(request.args.get('run_id', -1))
         played = request.args.get('played', False)
-        return render_with_data('play_finish.html', run_id=run_id, played=played)
+        return render_with_data('play_finish.html', type='sprint', run_id=run_id, played=played)
```
dqian3, line 145: I like having the sprint type here, but in the future we would likely want some way to figure this out on the frontend, so we can separate the frontend and backend.

templates/custom_play.html
```diff
@@ -0,0 +1,37 @@
...
+
+{% block content %}
+
+<div id="app">
+
+    <h2> Choose Your Starting and Ending Article! </h2>
+    <hr>
+
+    <div class="col-md-4 col-sm-6">
+        <ac-input :text="start" @update:text="newValue => start = newValue" placeholder="Start Article"></ac-input>
```
dqian3, line 17: I think this is where you can do the `v-bind:start.sync` thing? It's just shorthand for this as far as I know.

static/js/modules/autocomplete.js
```diff
@@ -0,0 +1,115 @@
+import { getAutoCompleteArticles } from "./wikipediaAPI/util.js";
```
dqian3, line 1: This looks great!

dqian3 (approved):

apis/runs_api.py
danielliu04: this is something that I don't know the best way to approach. So essentially I'm passing two strings through the url, and since there are spaces in the articles, i can't put them in directly. So i encoded it when passing in, and the unquote decodes it. its not the greatest, if there is a better way i can fix it

dqian3 (changes requested):
Actually, after trying it locally, you should integrate this mode into the front page changes that @maggarwal99 made. Make a new card/section for quick play/custom play. In fact, I think for custom play you can probably just include the form right there.

Also small nit, `/custom_play` has some vue artifacts upon initial load.

The autocomplete is awesome btw, great job on that. It seems really good on mobile too.

danielliu04:
Yeah I will work on that real quick before merging

scripts/schema.sql
```diff
@@ -190,4 +190,32 @@ CREATE TABLE IF NOT EXISTS `achievements_progress` (
     PRIMARY KEY (`achievement_id`, `user_id`),
     FOREIGN KEY (`achievement_id`) REFERENCES `list_of_achievements`(`achievement_id`),
     FOREIGN KEY (`user_id`) REFERENCES `users`(`user_id`)
-)
\ No newline at end of file
+);
```
bricehalder, line 193: Bump schema version in line 5 and in `db.py`. Trying to keep track of version changes with our db when we need to refer to past data.

bricehalder (changes requested):

scripts/schema.sql
danielliu04: Fixed

static/js/modules/autocomplete.js
```diff
@@ -0,0 +1,115 @@
...
+    data: function () {
+        return {
+            acList: [],
+            showAutocomplete: false,
+            highlightIndex: -1
+        }
+	},
+
+    methods: {
+        async setAutocomplete() {
```
dqian3, line 29: Another thought, how do you able to deal with multiple in flight requests returning at different times?

templates/generator.html
```diff
@@ -12,9 +12,13 @@
...
 
     data: {
-        start: "Start Article",
-        end: "End Article",
+        article: "Article",
+    },
 
+    methods: {
+        async generateRndPrompt() {
+            [ this.article ] = await this.$refs.pg.generatePrompt();
```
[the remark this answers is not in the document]
dqian3: I think the two-way binding was definitely a little complicated, but if we want to make this component completely untied from the input, an emitted event might be better.

views.py
```diff
@@ -1,6 +1,8 @@
 from cProfile import run
 from click import prompt
 from flask import Blueprint, render_template, request, redirect, session
+from toml import TomlDecodeError
```
dqian3, line 4: random import?

static/js/custom_play.js
```diff
@@ -0,0 +1,100 @@
...
+            articleCheckMessage: "",
+
+            scroll: null
+        }
+	},
+
+    methods: {
+
+        async generateRndPrompt(prompt) {
+            [this[prompt]] = await this.$refs.pg.generatePrompt();
```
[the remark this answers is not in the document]
maggarwal99: fancy js LOL

static/js/modules/game/articleRenderer.js
```diff
@@ -28,7 +28,12 @@ export class ArticleRenderer {
             // Create title
             let titleEl = document.createElement("div");
             titleEl.innerHTML = "<h1><i>" + body["title"] + "</i></h1>";
-            this.frame.insertBefore(titleEl, this.frame.firstChild);
+            this.frame.prepend(titleEl);
```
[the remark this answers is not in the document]
maggarwal99: I did but we should do that as part of the great refactoring

templates/generator.html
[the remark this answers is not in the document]
maggarwal99: A few reasons:
1. Following the single responsibility principle, we should make our components as small and simple as possible, i.e. do just 1 job. In the case of the generator, the only required parameter is the difficulty. We should keep all other templating/ui outside. Hence, why I removed the two-way binding on the start and end articles.
2. Considered using emit but eventually decided against it because with emit there's no way to bind contextually blindly i.e. how would you intercept what emit event is for which input group without coding them separately in the generator component.

static/js/modules/generator.js
```diff
@@ -12,66 +11,33 @@ async function getGeneratedPrompt(difficulty, num)
...
-                <div class="btn-group">
-                    <button class="btn btn-primary" v-on:click.prevent="generatePrompt"> Prompt (Start and End) </button>
-                    <button class="btn btn-primary" v-on:click.prevent="generateStart"> Start </button>
-                    <button class="btn btn-primary" v-on:click.prevent="generateEnd"> End </button>
-                </div>
-            </div>
-            <p>
-                <small><a href="/generator#about"> How does this work? </a></small>
-            </p>
+            <label class="form-label" style="width: 100%;">
```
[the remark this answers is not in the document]
maggarwal99: Not sure what you mean

dqian3 (approved):

PR #417, opened 2022-08-02
Per-prompt Stats on Leaderboards

bricehalder:
Adds metrics for completion rate, average play time (of completed runs), and average # of clicks (of completed runs) onto leaderboard pages for sprint prompts and lobby prompts. 

Dynamically updates depending on which leadership mode the user selects (uses all runs from whichever mode is selected).

Adds support for mobile viewing as well.

Test:

Basic view
Screen Shot 2022-08-02 at 6 13 01 PM

Changing leaderboard mode
Screen Shot 2022-08-02 at 6 13 46 PM

Mobile
Screen Shot 2022-08-02 at 6 14 29 PM

Lobby
Screen Shot 2022-08-02 at 7 17 44 PM

### DIFF
--- a/static/js/modules/wikipediaAPI/util.js
+++ b/static/js/modules/wikipediaAPI/util.js
@@ -71,4 +71,8 @@ async function getArticleSummary(page) {
     return body
 }
 
-export { getArticle, getArticleTitle, getArticleSummary, articleCheck };
+function round(value, decimals) {
+    return Number(Math.round(value+'e'+decimals)+'e-'+decimals);
+}   
+
+export { getArticle, getArticleTitle, getArticleSummary, articleCheck, round };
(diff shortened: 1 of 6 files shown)


static/js/modules/wikipediaAPI/util.js
```diff
@@ -71,4 +71,8 @@ async function getArticleSummary(page) {
     return body
 }
 
-export { getArticle, getArticleTitle, getArticleSummary, articleCheck };
\ No newline at end of file
+function round(value, decimals) {
```
dqian3, line 74: This shouldn't be in `wikipediaAPI/util.js`, these are meant to work with the wikipedia API, not our own. Also wow this is really the best way to round?

dqian3 (changes requested):
small change, sorry this took a while to look at

static/js/modules/wikipediaAPI/util.js
bricehalder: I don't know what I was doing here

PR #419, opened 2022-08-02
Updated stats_api

danielliu04:
Made all finished checks to be "finished" instead of "end_time IS NOT NULL".

Added new quick_runs statistics to the api.

Added "started_runs": runs where the user at least go to the loading page. If the first article page is not reached, the run data does not have a start time.

### DIFF
--- a/apis/stats_api.py
+++ b/apis/stats_api.py
@@ -81,7 +81,23 @@ def get_daily_stats():
         SUM(daily_plays) OVER (ORDER BY day) AS total 
     FROM data
     '''
-        
+    
+    queries['daily_started_sprints'] = '''
+    WITH data AS (
+        SELECT 
+            DATE(start_time) AS day,
+            COUNT(*) AS daily_plays 
+        FROM sprint_runs
+        GROUP BY day 
+    )
+
+    SELECT
+        day,
+        daily_plays,
+        SUM(daily_plays) OVER (ORDER BY day) AS total 
+    FROM data
+    '''
+
     queries['daily_finished_sprints'] = '''
     WITH data AS (
         SELECT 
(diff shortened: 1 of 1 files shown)


apis/stats_api.py
```diff
@@ -76,14 +81,30 @@ def get_daily_stats():
         SUM(daily_plays) OVER (ORDER BY day) AS total 
     FROM data
     '''
-        
+    
+    queries['daily_started_sprints'] = '''
```
bricehalder, line 85: I'm not really sure this gives us much more meaningful signal than what we already have in `daily_sprints` lines 68-83. Additionally, this won't do anything past schema v2.1 since we record `start_time` and `end_time` even if the user has not made it past the countdown screen. 

Screen Shot 2022-08-02 at 4 06 35 PM

top-to-bottom:
1. countdown button pressed @ 4s
2. countdown button pressed @ .9s
3. user exited during countdown @ .2s  (notice loadTime ≈ timeReached)
4. user exited during countdown @ 1s
5. user exited during coundown @ 4s

If we want to capture the amount of users that quit during countdown (which I think is the actual underlying stat this would uncover), I would take one of these approaches:

a) Change play_time to be an explicit 0 for users that did not make it past the countdown screen, and then query for that (+ `path` length of 1). I think @dqian3 and @mliu59 made the decision to still capture the countdown time as play_time, however.

b) Check for a `path` length of 1 and a difference between `loadTime` and `timeReached` < some epsilon.

c) Introduce a new field into our schema flagging if the user made it past countdown. 

Of these, option (c) feels the most robust, at the expense of adding another column. (extra ~265 kB as of writing this, so negligible).

apis/stats_api.py
```diff
@@ -104,6 +125,22 @@ def get_daily_stats():
...
     )
 
+    SELECT
+        day,
+        daily_plays,
+        SUM(daily_plays) OVER (ORDER BY day) AS total 
+    FROM data
+    '''
+
+    queries['daily_started_lobby_runs'] = '''
```
bricehalder, line 135: Same comment as above

bricehalder (changes requested):

apis/stats_api.py
danielliu04: Good point, I definitely got misleaded by some of the data in my database, so that is on me. I reverted the changes for now, and I will probably revisit this once we come up with a solution.

bricehalder:
re: NULLs in the database:

Screen Shot 2022-08-02 at 4 42 09 PM
We are seeing ~5000 NULLs in the database out of the past 115000 runs, so evidently there is still a significant portion of players that are hitting this. 

I suspect it's due to shoddy internet connection (not loading the first page in the background) or outdated browsers (i.e. not supporting `visibilitychange` or `onpagehide` or `unload` etc.). I can attempt a fix for this

bricehalder (approved):
looks good!

dqian3 (approved):

PR #427, opened 2022-08-10
Achievements change

danielliu04:
This is a refactor of achievements. Achievements is now a component and the code in html is a lot cleaner than before. 

One small addition is that there is a progress bar for multi-run achievements in the background.

Make sure to merge Quick-Play in first since this is based on the Quick-Play branch


dqian3 (approved):

static/js/play_finish.js
```diff
@@ -52,7 +58,9 @@ let app = new Vue({
         playTime: 0,
         loggedIn: false,
 
-        played: PLAYED
+        played: PLAYED,
+
+        isMounted: false
```
dqian3, line 63: For this, there's a handy directive called `v-cloak` that hides the template until vue is mounted.

static/js/play_finish.js
```diff
@@ -73,6 +81,11 @@ let app = new Vue({
         
         this.loggedIn = "username" in serverData;
 
+        // Make sure that a sprint can be considered for achievements
+        if (this.loggedIn) {
```
dqian3, line 85: What is this change for?

dqian3 (approved):

PR #429, opened 2022-08-10
Userlinks

danielliu04:
This creates a component for a user that can be used any place with a user displayed.

Right now this PR integrates into lobbies and the leaderboard.

### DIFF
--- a/static/js/modules/userDisplay.js
+++ b/static/js/modules/userDisplay.js
@@ -22,7 +22,7 @@ var userDisplay = {
     },
 
     template: (`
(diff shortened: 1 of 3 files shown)


static/js/modules/userDisplay.js
```diff
@@ -0,0 +1,29 @@
+
+var userDisplay = {
```
dqian3, line 2: small thing, but be consistent about naming components as PascalCase

dqian3: ehh I'll just do it for you

dqian3 (approved):

PR #450, opened 2022-10-02
recommend quickplay prompts

### LINKED ISSUE #444: Component that suggests quick play prompts from other users
Rather than having users rely on their own ideas or our generator to populate the quick play component, we could also have a list of prompts recently played (perhaps excluding randomly generated ones).

mliu59:
Adding a "recommended prompts" section to the quick play panel. On default loads the most recent 5 FINISHED quick runs and shows a play button for players to quickly jump into. 

image


config/default.json
```diff
@@ -2,8 +2,8 @@
     "SECRET_KEY": "ahhhhhhhhhh",
 
     "MYSQL_HOST": "127.0.0.1",
-    "MYSQL_USER": "user",
-    "MYSQL_PASSWORD": null,
+    "MYSQL_USER": "root",
```
dqian3, line 5: oops, put your local stuff in `config/prod.json`

dqian3 (changes requested):
Small change. Also I would say that these shouldn't be hidden completely in a `<details>` element. Maybe show 1 or 2 and an option to see more. Also would be nice to see the time it was played, i.e. using `moment.js`

bricehalder:
+ this would be an easy target to grief i.e. putting nsfw stuff in the quick plays box

mliu59:
> plus this would be an easy target to grief i.e. putting nsfw stuff in the quick plays box.
> 
> Maybe instead of most recent, we could do something like most played in the past day or something?

I changed the suggested prompts to be random finished prompts from the past 3 days, so griefing would be practically unviable with sufficient traffic volume. I think griefing would be easier to do if it was just most played since they can just spam.

dqian3 (approved):

PR #467, opened 2022-11-28
Remove ability to view other prompts' runs on leaderboard + slight refactor

### LINKED ISSUE #215: Leaderboard with non-matching prompt and run URL parameters displays incorrectly
When calling http://127.0.0.1:5000/prompt/3?run_id=2, in the database run_id 2 corresponds with prompt_id 1, but calling it with prompt_id 3 returns results from both prompt_id 2 and prompt_id 3 instead of the expected behavior of prompt_id 3 or an error message.

Screen Shot 2022-02-27 at 2 34 38 AM

bricehalder:
Fixes #215, previously you could just switch the run_id in the leaderboard and it would display the run regardless of which prompt leaderboard you are viewing. e.g. something like this https://wikispeedruns.com/leaderboard/399?preset=ffa&offset=0&run_id=101

With this change, we simply check to see if the current run (run_id) is actually part of the prompt that is being viewed. 

Also cleaned up the leaderboard api a bit.

Test:
Before:
Screen Shot 2022-11-27 at 10 55 36 PM
(Another prompt's run is shown)

After:
Screen Shot 2022-11-27 at 10 56 00 PM
(Run is not shown)

Verified that leaderboard sort modes still work.

### DIFF
--- a/wikispeedruns/leaderboards.py
+++ b/wikispeedruns/leaderboards.py
@@ -25,6 +25,51 @@ def get_db():
     return conn
 '''
 
+LEADERBOARD_COLUMNS = ['run_id',
+        'start_time',
+        'end_time',
+        'play_time',
+        'finished',
+        'path',
+        'prompt_id',
+        'user_id',
+        'username',
+        'rank',
+        'numRuns',
+        'path_length']
+
+BASE_ALIAS = 'runs'
+OUTER_ALIAS = 'processed_runs'
+
+def _query_select_columns(table):
+    separator = ', '
+    if __debug__:
+        separator = ',\n\t'
+
+    return separator.join([f'{table}.{column}' for column in LEADERBOARD_COLUMNS])
+
+def _query_current_run_clause(table, run_id=None):
+    if run_id is None:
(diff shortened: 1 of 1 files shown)


wikispeedruns/leaderboards.py
```diff
@@ -25,6 +25,51 @@ def get_db():
...
+    if __debug__:
+        separator = ',\n\t'
+
+    return separator.join([f'{table}.{column}' for column in LEADERBOARD_COLUMNS])
+
+def _query_current_run_clause(table, run_id=None):
+    if run_id is None:
+        return '0'
+
+    return f'({table}.run_id = %(run_id)s AND {table}.prompt_id = %(prompt_id)s)'
```
bricehalder, line 55: The only practical change in this diff comes from this line, where the change is
`current_run_clause = 'runs.run_id = %(run_id)s'` 
to
`current_run_clause = '(runs.run_id = %(run_id)s' AND runs.prompt_id = %(prompt_id)s)`

dqian3 (approved):

PR #469, opened 2022-11-30
Loading timer on pulling new stats, fix date shown

bricehalder:
Adds an incrementing counter while we are polling for stats.

Also always use utc time so it works regardless of device time. 

Screen Shot 2022-11-30 at 1 08 05 AM

### DIFF
--- a/apis/stats_api.py
+++ b/apis/stats_api.py
@@ -25,7 +25,15 @@ def calculate_stats():
 @stats_api.get("/all")
 @check_admin
 def get_total_stats():
-    most_recent_stats_query = 'SELECT * FROM `computed_stats` ORDER BY `timestamp` DESC LIMIT 1'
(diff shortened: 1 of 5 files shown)


apis/stats_api.py
```diff
@@ -25,7 +25,15 @@ def calculate_stats():
 @stats_api.get("/all")
 @check_admin
 def get_total_stats():
-    most_recent_stats_query = 'SELECT * FROM `computed_stats` ORDER BY `timestamp` DESC LIMIT 1'
+    most_recent_stats_query = """
+    SELECT 
+        stats_json, 
+        DATE_FORMAT(timestamp, '%Y-%m-%dT%TZ') AS timestamp
```
bricehalder, line 31: give timestamp in ISO format

dqian3 (approved):

PR #470, opened 2022-11-30
Add error handling for database version

bricehalder:
This would throw an error anytime db is upgraded

### DIFF
--- a/wikispeedruns/achievements/achievements.py
+++ b/wikispeedruns/achievements/achievements.py
@@ -35,7 +35,8 @@ def convert_to_standard(raw_data: Dict[str, Any]) -> Dict[str, Any]:
     version_map = get_version_map()
     path = json.loads(raw_data["path"])
     version = path["version"]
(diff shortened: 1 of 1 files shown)


wikispeedruns/achievements/achievements.py
```diff
@@ -35,7 +35,8 @@ def convert_to_standard(raw_data: Dict[str, Any]) -> Dict[str, Any]:
     version_map = get_version_map()
     path = json.loads(raw_data["path"])
     version = path["version"]
-    return version_map[version](raw_data)
+    version_func = version_map[version] if version in version_map else version_2_1
```
dqian3, line 38: Why not add the new version to the map?

bricehalder: I can if you'd like, but this will always be an issue on each db version bump. Is there a good reason we shouldn't just use this as the default (until the default needs to be changed)? @danielliu04

dqian3 (approved):

PR #479, opened 2023-01-25
Use revision API in POTD to curb wiki vandalism

dqian3:
This change uses the release date of the Prompt of the day (i.e. `active_start`) as a date to query revisions from wikipedia from. Specifically, we take the most recent query before `active_start` datetime. This applies to all articles in a sprint. If we want to add the same functionality to lobby prompts, we would need a creation datetime.

I also cleaned up some old announcements and added a new one shaming people for editing wikipedia.

As an example, the follow page had some [edits ](https://en.wikipedia.org/w/index.php?title=Witold_Lutos%C5%82awski&diff=1135506581&oldid=1133531527) on Jan 25:
 
image

In my local instance, I added a prompt released on Jan 23 and the edits are gone:

image

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -39,11 +39,13 @@
     <div>
         <h3 class="mb-4">Welcome<span v-cloak v-if="username"> back</span> to Wikipedia Speedruns<span  v-cloak v-if="username">, [[username]]</span>!</h3>
 
-        <div class="alert alert-warning" v-cloak role="alert" v-html="streakText" v-if="loggedIn && streakText"></div>
-
-        <div class="alert alert-info" v-cloak role="alert" v-if="loggedIn">
-            We've just added <b>achievements</b> to the game! Visit your profile page <a v-bind:href="'/profile/' + username">here</a> to check your current progress.
+        <!---->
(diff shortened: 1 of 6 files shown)


templates/home.html
```diff
@@ -39,12 +39,14 @@
     <div>
         <h3 class="mb-4">Welcome<span v-cloak v-if="username"> back</span> to Wikipedia Speedruns<span  v-cloak v-if="username">, [[username]]</span>!</h3>
 
-        <div class="alert alert-warning" v-cloak role="alert" v-html="streakText" v-if="loggedIn && streakText"></div>
-
-        <div class="alert alert-info" v-cloak role="alert" v-if="loggedIn">
-            We've just added <b>achievements</b> to the game! Visit your profile page <a v-bind:href="'/profile/' + username">here</a> to check your current progress.
+        <!---->
+        <div class="alert alert-info" v-cloak role="alert" v-if="new Date() < new Date('2023-01-31')">
```
maggarwal99, line 43: Don't think we need to add an alert for this, draws unnecessary attention imo. Can add it to the game rules instead. Although, calling out that our game isn't affected by edits discourages mal edits which is what we want so maybe it's fine.

maggarwal99 (approved):

PR #484, opened 2023-04-06
Community submitted prompts

mliu59:
PR includes the fullstack for community submitted prompts, including:

- Users submit prompts into a pending prompt pool
- Admins approve/reject pending prompts, adding them to pool of prompts ready to be used as POTDs
- Showing credit for community submitted prompts.


dqian3 (commented):
This is awesome! Glad to see some new work being done. Just had the few comments about the schema and a few higher level comments here:
1. Could you post some screenshots when you have the complete PR?
2. Would love to see marathon prompts, though that would be a lot of UI work I assume?
3. What do you think about putting the generator with the community add? Do we want to encourage people touse it?

mliu59:
> 3. What do you think about putting the generator with the community add? Do we want to encourage people to use it?

I think this will definitely be helpful to have. Depending on the quality and quantity of prompts we get, we may or may not need some auto-filtering step before something gets presenting to admins for approval. Auto filtering can potentially catch any prompts from the generator that's too easy

dqian3:
I guess I just don't want people to submit generated prompts straight from the generator; we can do that ourselves.

mliu59:
I personally think that using a prompt generator should be allowed, and if we get enough volume of community prompts in the future admins might never need to personally come up with prompts anymore, with or without the generator. 

I can definitely see how allowing the generator may cause some spamming. My guess is that users that participate in submitting prompts will either never use the generator, or almost always use the generator. Having the approval queue that interleaves between prompts submitted by different users can maybe mitigate this issue a bit.

mliu59:
Some screenshots of the most recent GUI commit
Sprints:
image
image

Marathon: 
image
image

Prompt builder in the admin pages do not show the anonymous selection, and marathon will have a "Load example" button for testing functionality
image

mliu59:
Additional improvements:
- [ ] Evaluate this page on mobile. Prompt builders, especially marathon, can be very cluttered on mobile. 
- [ ] Redirect users to login page if not logged in upon landing on this page. We don't want a user not logged in to spend time building a prompt only to find out they are not logged in after attempting to submit

mliu59:
Screenshots for admin approval UI
Sprints: admins have the option to approve, reject, or approve as anonymous. List will update after every action, and will only show the first 10 on the approval queue. 
image

Marathon: Approval process is the same as sprints
image

mliu59:
Showing creator of prompts (WS team vs. user vs. anonymous user)
image
image

On countdown:
image
On leaderboard:
image

For prompts created by admins
image
vs. anonymous submission
image

dqian3:
Look great so far! I think one small thing I'd want is for the buttons on the marathon UI to be bootstrap buttons, but other than that looks great! are there links on the home page? Trying to access on beta

dqian3:
ok, playing around with it more, a few comments. 
- In addition to the feedback you mentioned, I would want something like a list of submitted prompts, along with checkmarks or something to show that they got approved and any that have been played in the past
- (small) the marathon UI allows for adding more than 5 starting prompts. What happens then?
- Add an achievement for submitting a prompt! Lol

mliu59:
> * [ ]  (small) the marathon UI allows for adding more than 5 starting prompts. What happens then?

Currently, we allow more than 5 starting prompts in the builder, but there can only be 5 when submitting. Error msg will show, and the prompt is not cleared. 
image

mliu59:
Added an additional section in the cmty page that shows a user's approved and pending prompts. 
image
image

bricehalder:
Looks really good overall!!! Some design feedback:

IMO the creator doesn't need to be prominently featured as it is now - let's just say there are a lot of really weird/inappropriate usernames lol. Having it be the same size makes the screen more cluttered and it's not as important as the other stuff so it should not take up the same real estate.
  * On the home screen, I think a distinction between Official prompts and Community prompts is good enough
  * Indifferent about it on the countdown screen, but I think if it remains it should be more minimal compared to the rest. Suggestions:
     * Smaller/thinner font
     * Placed below the countdown/bottom of the screen
  * I think the leaderboard _is_ an appropriate place to have it, but I think it should be less prominent. Underneath the start and end articles makes the most sense to me. Alternatively, we could move the start/end to the first line i.e. 

> **Prompt 55:** **Geographic coordinate system** to **England**
(small) Created by: **admin**

scripts/schema.sql
bricehalder, line 1: Make sure to bump the schema version in this file and in db.py

mliu59: Updated

mliu59:
> IMO the creator doesn't need to be prominently featured as it is now - let's just say there are a lot of really weird/inappropriate usernames lol. Having it be the same size makes the screen more cluttered and it's not as important as the other stuff so it should not take up the same real estate.

Agreed here. Previously when there wasn't a section where users can check on the status/popularity of their submitted/pending prompts, I felt that a more prominent feature is necessary. However, now with that implemented, decluttering UI is the better option. 

> * On the home screen, I think a distinction between Official prompts and Community prompts is good enough

Implemented and reduced font size

> * Indifferent about it on the countdown screen, but I think if it remains it should be more minimal compared to the rest. 

Opted to completely remove it from countdown. The countdown UI is supposed to provide as much info about the upcoming prompt as possible. Adding creator credits only distracts/clutters. 

> * I think the leaderboard _is_ an appropriate place to have it, but I think it should be less prominent. Underneath the start and end articles makes the most sense to me. Alternatively, we could move the start/end to the first line i.e.

Moved under start/end

image
image

dqian3:
Looks great! One last ask from me. I think there should be some permanent links to the community suggestion page on the home page. The banner is to highlight new features. Also I think the URL should be something more intuitive than `/cmty`. Maybe `/suggest` or `contribute`.

mliu59:
Reframing /cmty to be more of a general "contribute" page. Main focus is still on submitting prompts, but I also included footer section with links to our discord, github, and donor box :) Because now this contains more community information, I opted to include the link in the nav menu for both logged in and logged out users. 
image
image

Logged in users can see the prompt builder section. Not logged in users won't
image
image

Open to all feedback

dqian3 (approved):

PR #486, opened 2023-04-14
Forbid users from searching with Ctrl+G

VinceUB:
In Firefox and Chrome (and probably other browsers), you can search a page with Ctrl+G, and this is not detected by wikispeedruns. 

I believe that I have fixed this.


bricehalder:
Thanks!!! We should really reduce our code duplication

bricehalder (approved):

Issue #393, opened 2022-07-06
Add option to remove prompts with runs


mliu59:
Can be used to permanently remove bad prompts that users report

Issue #381, opened 2022-06-19
Add manage option to remove prompt from POTD without deleting


mliu59:
I dont think its currently possible to remove a prompt from POTD without deleting it entirely, which isn't possible if there are already runs for it. 

This would be used in situations such as:
- Players report that a prompt is impossible
- Accidentally adding an old prompt as a recent POTD

PR #491, opened 2023-05-02
various admin QOL/management tools

mliu59:
Added features so far:
- button to move a prompt assigned as POTD already to unused. Currently only working for prompts in the two week window #381 
- allow admins do delete prompts with existing runs. A popup will inform admin that a prompt already has runs, admin will then need to confirm. Deletes all associated runs #393 
- Auto populate POTD button #492


bricehalder (approved):
thanks for doing this!!

PR #494, opened 2023-05-03
cleaned up home page text, added about to navbar

mliu59:
PR for cleaning up our home page UI, which is pretty cluttered. 
image

- Less intro text
- Added about to navbar

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -51,18 +51,20 @@
 
         <div class="alert alert-warning" v-cloak role="alert" v-html="streakText" v-if="loggedIn && streakText"></div>
 
-        <p>🏁 <b>Explore the world's largest encyclopedia with lightning speed!</b> Navigate from a starting Wikipedia article to the target using only links shown in each page as fast as you can. </p>
+        <p>🏁 <b>Explore the world's largest encyclopedia with lightning speed!</b> Start from one Wikipedia page and navigate to the goal page as fast as you can!</p>
             
-        <p>First time? Play the <a href="/play/tutorial">tutorial</a>!</p>
+        <p>Remember, luck is king, so put on your lucky hat, grab your lucky socks, and let the games begin!</p>
 
-        <p>Try our new prompt generator <a href="/generator">here</a>!</p>
+        <p>First time? Play the <a href="/play/tutorial">tutorial</a></p>
+
+        <p>Try our new prompt generator <a href="/generator">here</a></p>
 
         <template v-cloak>
             <p v-if="loggedIn === true && lobbies.length > 0">
                 See your active lobbies <a href="#user-lobbies">here</a> or <a href="/lobby/create">create a new private lobby</a>.
             </p>
             <p v-if="loggedIn === true && lobbies.length == 0">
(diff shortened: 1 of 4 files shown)


templates/home.html
```diff
@@ -51,14 +51,8 @@ <h3 class="mb-4">Welcome<span v-cloak v-if="username"> back</span> to Wikipedia
...
 
-        <p>The rules of this game are simple. You'll be given a starting Wikipedia article and an ending Wikipedia article. Your goal is to navigate from the starting article to the ending article using only the article links.</p>
-        <p>Some other things to keep in mind:</p>
-        <ul>
-            <li>Any link (outside the navigation bar at the top) is fair game.</li>
-            <li>You can't use the browser find tool.</li>
-            <li>Reloading the page or clicking the browser back button will end your run.</li>
-        </ul>
-        <p>Remember, luck is king, so put on your lucky hat, grab your lucky socks, and let the games begin!</p>
+        <p>🏁 <b>Explore the world's largest encyclopedia with lightning speed!</b> Navigate from a starting Wikipedia article to the target using only links shown in each page as fast as you can. </p>
```
bricehalder, line 54: Hmm I think this is still a little overcomplicated, I think

"Start from one Wikipedia page and navigate to the goal page as fast as you can!"

1) target --> goal is clearer I think
2) Simpler is better + the part about clicking links is kind of confusing, so it does more harm than good imo. It will be obvious how to play for most people, and for those who are confused we have the tutorial that is very explicit.

templates/home.html
```diff
@@ -51,14 +51,8 @@ <h3 class="mb-4">Welcome<span v-cloak v-if="username"> back</span> to Wikipedia
...
         <div class="alert alert-warning" v-cloak role="alert" v-html="streakText" v-if="loggedIn && streakText"></div>
 
-        <p>The rules of this game are simple. You'll be given a starting Wikipedia article and an ending Wikipedia article. Your goal is to navigate from the starting article to the ending article using only the article links.</p>
-        <p>Some other things to keep in mind:</p>
-        <ul>
-            <li>Any link (outside the navigation bar at the top) is fair game.</li>
-            <li>You can't use the browser find tool.</li>
-            <li>Reloading the page or clicking the browser back button will end your run.</li>
-        </ul>
-        <p>Remember, luck is king, so put on your lucky hat, grab your lucky socks, and let the games begin!</p>
```
bricehalder, line 61: :(  line 61 gave our site some personality but I'm fine deleting lol

bricehalder (approved):

bricehalder:
Accepted but I did have a comment about changing some of the text

templates/home.html
mliu59: Agreed, I'll add in this change

templates/home.html
mliu59: Yeah I really wanted to keep this line as well, but having just one sentence as a tag line does make it alot less cluttered. Will experiment with this a bit more.

mliu59: Opting to keep this line.

PR #495, opened 2023-05-03
:%s/Wikispeedruns/WikiSpeedruns/g

bricehalder:
Wikipedia Speedruns, Wiki Speedruns, or WikiSpeedruns are probably all fine but our dev blog uses WikiSpeedruns so using that. Personally I prefer Wiki Speedruns (natural shortening from Wikipedia Speedruns and is most clear) but they are both better than Wikispeedruns


mliu59 (approved):
Looks good!

dqian3 (approved):

PR #496, opened 2023-05-05
Prompt submission hotfix

mliu59:
Sprint builder will use CheckArticles to get redirect/actual article titles instead of directly submitting user text input.


bricehalder (approved):
nice

PR #508, opened 2023-08-23
Add end article when share result

ant38:
When we click on "Share run summary", only start article is copied to clipboard. This update add also end article.
From:
```
WikiSpeedruns
United States
4 🖱️
16.687 ⏱️
https://wikispeedruns.com/play/quick_play?prompt_start=United%20States&prompt_end=India&lang=en
```
To:
```
WikiSpeedruns
United States > India
4 🖱️
16.687 ⏱️
https://wikispeedruns.com/play/quick_play?prompt_start=United%20States&prompt_end=India&lang=en
```

### DIFF
--- a/static/js/pages/play_finish.js
+++ b/static/js/pages/play_finish.js
@@ -168,7 +168,7 @@ let app = new Vue({
 
 
         generateResults: function(event) {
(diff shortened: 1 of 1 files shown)


bricehalder:
I'd be fine adding this just for Quick Runs (so this PR would have to be slightly modified), what do y'all think @mliu59 @dqian3

dqian3:
Yeah, I agree, @ant38 if you could make it only for quick runs, that would be great

dqian3 (approved):

PR #511, opened 2023-10-05
added "created by" col to archive

mliu59:
image


bricehalder (approved):
Want to link to their profile page like we do in prompts?

Also I forget if we use "WikiSpeedruns Team" or "WS Team", but we should be consistent. I think I like "WikiSpeedruns Team" more since it's more explicit

PR #517, opened 2023-11-19
Delete lobby

kwonghua:
added functionality to delete lobby from lobby page

### DIFF
--- a/templates/leaderboard.html
+++ b/templates/leaderboard.html
@@ -36,7 +36,7 @@
                             <strong>[[prompt.start]]</strong> to <strong v-if="runs.length > 0">[[prompt.end]]</strong><strong v-else>...</strong>
                         </h6>
                         
-                        <small> Prompt created by: 
+                        <small v-if="!lobbyId"> Prompt created by: 
                             <template v-if="prompt.username && prompt.cmty_anonymous">
                                 <strong>Anonymous User</strong>
                             </template>
--- a/wikispeedruns/lobbys.py
+++ b/wikispeedruns/lobbys.py
@@ -375,7 +375,8 @@ def get_user_lobbys(user_id: int):
         active_date,
         rules,
         user_lobbys.owner,
-        count(prompt_id) as n_prompts
+        count(prompt_id) as n_prompts,
+        isHidden
     FROM lobbys
     LEFT JOIN user_lobbys ON user_lobbys.lobby_id=lobbys.lobby_id
     LEFT JOIN lobby_prompts ON lobby_prompts.lobby_id=lobbys.lobby_id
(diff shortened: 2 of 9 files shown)


wikispeedruns/lobbys.py
```diff
@@ -453,4 +453,24 @@ def get_lobby_anon_users(lobby_id: int):
...
+def delete_lobby(lobby_id) -> bool:
+    setCheckOff = """
+    SET FOREIGN_KEY_CHECKS=OFF
+    """
+    query = """
+    DELETE FROM lobbys 
+    WHERE lobby_id=%s
+    """
+    setCheckOn= """
+    SET FOREIGN_KEY_CHECKS=ON
```
bricehalder, line 468: This is a great start!!!! This change is a little complicated because we need to decide how we want to handle deleted lobbies. Currently this change would leave a lot of dangling references to deleted `lobby_id`s in `lobby_prompts` and `lobby_runs`. Writing out questions for myself/WikiSpeedruns Team.

We should probably delete all corresponding entires in the `user_lobbys` table. After that: 

Some options I can think of - 
1) Add a new `isHidden` column to `lobbys` table and just use that, no fiddling with `lobby_id`s
2) Set `lobby_id` to NULL in `lobbys` and/or `lobby_runs` and `lobby_prompts`
3) Delete entries from `lobbys` and corresponding runs and prompts from `lobby_runs` and `lobby_prompts`
4) Current implementation of just deleting entries from `lobbys` (in addition to `user_lobbys` as stated above) and just be fine with / handle dangling references
5) Some other combination of the above three or other ideas


@dqian3 @mliu59 thoughts?

bricehalder (changes requested):

bricehalder:
Good start, also would be good to upload screenshot or better a video to show usage / testing the change

wikispeedruns/lobbys.py
dqian3: I like option 1, it's the easiest and keeps things from getting messed up. Furthermore, we don't have to worry about consistency issues if we add features to lobbies.

I think there's also some SQL magic to make this work and delete references (DELETE CASCADE), which could be another option.

templates/leaderboard.html
```diff
@@ -36,7 +36,7 @@ <h6 v-if="available">
                             <strong>[[prompt.start]]</strong> to <strong v-if="runs.length > 0">[[prompt.end]]</strong><strong v-else>...</strong>
                         </h6>
                         
-                        <small> Prompt created by: 
+                        <small v-if="!lobbyId"> Prompt created by:
```
mliu59, line 39: I think this was added in #510?

wikispeedruns/lobbys.py
mliu59: I think setting a lobby to hidden would be a good idea It'll also allow us to keep metrics related to lobby creation/deletion (what types of lobbies get deleted, etc.)

Resonantjacket5: I'd suggest we can probably keep most of this delete code for now just submitting the lobbys_api.py changes but without the delete button. and then have a separate pr renaming the button to 'archive' and set the lobby to hidden there, this also sets users expectations clearer on what is happening.

bricehalder: > fancy SQL magic

Yeah we can do ON DELETE CASCADE to delete rows in other tables that reference the lobby or ON DELETE SET NULL to just null out the column in other tables

templates/leaderboard.html
bricehalder: Yeah, this will merge cleanly so I think it's fine

kwonghua:
I made some changes based on the comments above. 

As for handling the deletion of lobbies, I went with the recommended approach and added a `isHidden` column in the `lobbys` table.  
However with that we would need to run when merging to prod:
`ALTER TABLE lobbys ADD isHidden BOOLEAN DEFAULT 0;` 

Currently there is a minor issue in the case of if all lobbies are marked as hidden then on the home page in the lobby section under Party Mode, the default message
`You don’t have any existing lobbies.` will not display.

bricehalder:
LGTM, @dqian3 for merging to prod

dqian3 (approved):

PR #528, opened 2023-11-21
Add dark mode

Resonantjacket5:
Changes
* Updates bootstrap from 5.1.3 to 5.3.2 (For themes)
* Add switch theme button to nav bar. 
* Modified nav.js with switchTheme function and saveTheme (save's the theme locally as a setting). Setting is saved locally using localStorage (going to another computer does not persist the setting).
* Add wikipedia mobile css override file to replace background-color and font color with bootstrap variables (mobile files are referenced externally) 
* Modified wikipedia desktop css file (we have those saved locally) to replace background-color and font color with bootstrap variables

Other notes
* Tried using media preference (prefers-color-scheme) for inferring theme but didn't seem to work properly and no one has that setting anyways on desktop.
* Maybe in the future could be saved server side in sql table.
* For inverting colors, another alternative was to use "filter: invert(1) hue-rotate(180deg)!important;" on the wiki-insert div but ran into trouble since currently partially use bootstrap css in addition to the wikipedia css so could not cleanly invert all colors.
* Tried adding button as a toggle or sun/moon symbol but ran into trouble with jinga/vue maybe for next pr

### DIFF
--- a/static/stylesheets/main.css
+++ b/static/stylesheets/main.css
@@ -73,7 +73,7 @@ a:hover, a:hover * {
     display: inline-block;
     min-width: 40%;
     width:auto;
-    background-color: #F8F8F8;
+    background-color: var(--bs-body-bg);
     border: 2px solid #000000 !important;
 
     height: auto;
@@ -89,7 +89,7 @@ a:hover, a:hover * {
     width: 100vw;
     bottom: 0;
     left:0;
-    background-color: #F8F8F8;
+    background-color: var(--bs-body-bg);
     border-top: 3px solid #000000 !important;
 
     padding: 5px 10px;
--- a/static/stylesheets/main.css
+++ b/static/stylesheets/main.css
@@ -66,7 +66,7 @@ a:hover, a:hover * {
 }
 
 .HUDwrapper-fade {
-    background-image: linear-gradient(#ffffff00, #ffffffbb, #ffffffbb, #ffffff);
+    background-image: linear-gradient(#ffffff00, rgba(var(--bs-body-bg), 0.8), var(--bs-body-bg));
 }
 
 .HUD {
--- a/templates/base.html
+++ b/templates/base.html
@@ -74,8 +74,9 @@
                         </template>
 
                         <template v-else>
-                            <a class="btn btn-light" href="/register">Register</a>
-                            <a class="btn btn-light" href="/login">Login</a>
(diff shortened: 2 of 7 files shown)


bricehalder:
This is epic!! Thanks for this awesome effort. Can you upload screenshots or a video to show its usage?

Resonantjacket5:
Had to lower quality and trim to fit 10 megabyte limit.

https://github.com/wikispeedruns/wikipedia-speedruns/assets/5826282/61a7623e-0e48-42d3-925f-0ba77f385061


https://github.com/wikispeedruns/wikipedia-speedruns/assets/5826282/1d169eb6-c8f5-4efa-bd7d-577e63081686

templates/base.html
```diff
@@ -55,7 +55,9 @@
                         <li class="nav-item mx-2">
                             <a class="nav-link" v-bind:class="loggedIn ? '': 'text-center'" v-bind:href="'/contribute'">Contribute</a>
                         </li>
-
+                        <button class="btn" v-on:click="switchTheme()">
+                            SwitchTheme
```
dqian3, line 59: Actually, maybe this could be more informative. I'll do that myself actually.

dqian3:
Dang, this is what happens when someone who knows what they're doing with CSS looks at our site lol. Will merge with the batch of other changes!

static/stylesheets/main.css
```diff
@@ -66,14 +66,14 @@ a:hover, a:hover * {
 }
 
 .HUDwrapper-fade {
-    background-image: linear-gradient(#ffffff00, #ffffffbb, #ffffffbb, #ffffff);
+    background-image: linear-gradient(#ffffff00, rgba(var(--bs-body-bg), 0.8), var(--bs-body-bg));
```
dqian3, line 69: Changed this, cause we got a white bar at the bottom of the page.

dqian3:
> Ohhhh I see we’ve changed the delimiters from the {{}} to [[]]. :)
> 
> I thought it was jinga messing it up even after adding escape characters.

Oh yeah, you're actually right that it was jinja, with our weird setup of flask serving vue templates, ``{{}}`` were used by both jinja and vue, and so conflicted. So we changed the vue delimiter to ``[[]]``

dqian3:
@bricehalder @mliu59 Could one of you go through and check if there's any other custom components that we missed. I think it should be good

mliu59:
merge merge merge

PR #531, opened 2023-11-27
Revert "recommend quickplay prompts"

dqian3:
Reverts wikispeedruns/wikipedia-speedruns#450


dqian3:
It seems like the query in #450 was taking too long, and would cause issues...

PR #532, opened 2023-11-27
Revert "Revert "recommend quickplay prompts"" (i.e. original PR changes)

dqian3:
Reverts wikispeedruns/wikipedia-speedruns#531

Originally https://github.com/wikispeedruns/wikipedia-speedruns/pull/450


dqian3:
There was an issue with this PR where the query to get recommended quick_play prompts would take too long and timeout, causing issues on the entire WS server. Might be worth revisiting by adding a SQL `LIMIT`

PR #533, opened 2023-11-27
Recommend Quickplay Prompts

dqian3:
Reopening #450 after reverting it in #531

There were issues with the SQL query taking too long and timing out, might be worth taking a look into using a `LIMIT` or something

### DIFF
--- a/apis/runs_api.py
+++ b/apis/runs_api.py
@@ -9,6 +9,8 @@ import json
 from datetime import datetime
 from wikispeedruns import prompts, lobbys, runs
 
+import random
+
 # Does not have its own prefix, instead part of the lobbys and sprints apis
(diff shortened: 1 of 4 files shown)


Resonantjacket5:
This is probably doing a table scan since there is no secondary index on start_time /end _time?

Though I can run sql analyze later and double check.

mliu59:
@dqian3 @Resonantjacket5 We def don't have a secondary index on the timestamps, and I didn't think we needed to add one just for this feature, so I changed the query to take the most recent 300 completed runs using run_id, and randomly select those with unique start, end, and user_id combinations. This should** be a lot faster. 

I also added some quick_run entries to populate_db.py

dqian3:
Let's try it lol

PR #534, opened 2023-11-27
Fixed the hidden lobby bug

kwonghua:
empty lobby message will now display correctly to users in the case of all hidden lobbies.

### DIFF
--- a/static/js/pages/home.js
+++ b/static/js/pages/home.js
@@ -98,11 +98,7 @@ var app = new Vue({
         this.activePrompts = prompts.filter(p => !p.rated);
 
         this.lobbies = await getUserLobby(this.loggedIn);
-        for(let lobby of this.lobbies){
-            if(lobby.isHidden){
-                this.hiddenLobbies++;
-            }
-        }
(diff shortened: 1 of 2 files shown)


static/js/pages/home.js
```diff
@@ -98,6 +98,11 @@ var app = new Vue({
         this.activePrompts = prompts.filter(p => !p.rated);
 
         this.lobbies = await getUserLobby(this.loggedIn);
+        for(let lobby of this.lobbies){
```
bricehalder, line 101: If you want to modernize / functionalize this you can do something like 
```
this.hiddenLobbies = this.lobbies.reduce((count, lobby) => count + lobby.isHidden, 0);
```

bricehalder (approved):
Nice, thanks for quickly fixing this!!!

static/js/pages/home.js
bricehalder: If you want to make the change and commit it feel free and then we can merge this. Here is some reading on map, reduce, and filter in JS

kwonghua: Gotcha, thanks for that link too. I went ahead and made the changes suggested

PR #538, opened 2023-12-01
Adding a swap button for quickplay and sprint builder

### LINKED ISSUE #527: Add button to swap quick play prompts
Screenshot 2023-11-20 at 11 41 17 PM

Sometimes you want an article as the start/end or vice versa. Easier to have a button rather than to re-type things. QOL change. Would be really useful for Sprint Builder in admin tools as well.

mliu59:
the button was surprisingly annoying to add, but it is here

image
image
 on mobile:
image


bricehalder (approved):
Awesome, this will make my life a lot easier when adding prompts! I think the only thing is the mobile UI. Maybe have it be vertically in-between instead of smushing the end article. Or just not show it at all for mobile if it's too complicated.

mliu59:
> Awesome, this will make my life a lot easier when adding prompts! I think the only thing is the mobile UI. Maybe have it be vertically in-between instead of smushing the end article. Or just not show it at all for mobile if it's too complicated.

Removed swap button for mobile.

mliu59:
oops added commits to the wrong branch. Reverted

dqian3:
LGTM

PR #545, opened 2023-12-06
added an archives page for marathon

### LINKED ISSUE #544: Create an archives page for marathons + reduce featured number of marathons on home page
One of the biggest barriers to accepting/creating more marathons is the amount of screen space it takes on the home page. We'll need to create another archives page for it.

mliu59:
- added an archives page for marathons, nearly a carbon copy of the sprints archive. It also uses the limit/offset pagination. Default limit of prompts on a page of the archive is 10 (as opposed to the 20 for sprints)
- reduced featured list size on home page to 5. It will only show the most recent 5 (sorted by prompt id)
image
- marathon archive will also check for saved games. If there is a saved game for a prompt, the play link will be "continue" to continue saved game instead. 
image
- Mobile looks the same as the home menu table.

### DIFF
--- a/apis/marathon_api.py
+++ b/apis/marathon_api.py
@@ -124,26 +124,16 @@ def get_all_marathon_prompts():
 
 
 @marathon_api.get('/get_prompts')
-def get_marathon_prompts():
+def get_marathon_prompts():    
+    try:
+        limit = int(request.args.get('limit', 5))
+        marathons, _ = prompts.get_archive_prompts("marathon",
+            limit=limit)
 
-    limit = int(request.args.get('limit', 5))
+        return jsonify(marathons)
 
-    args = {'p_limit': limit}
-
-    query = """
-    SELECT prompt_id, start, initcheckpoints, checkpoints, cmty_added_by, cmty_anonymous, cmty_submitted_time, username FROM marathonprompts
-    LEFT JOIN users
-    ON users.user_id = marathonprompts.cmty_added_by
-    ORDER BY prompt_id DESC
-    LIMIT %(p_limit)s
-    """
-
-    db = get_db()
-    with db.cursor(cursor=DictCursor) as cursor:
-        cursor.execute(query, args)
-        results = cursor.fetchall()
-
-        return jsonify(results)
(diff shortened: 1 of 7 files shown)


wikispeedruns/prompts.py
```diff
@@ -249,12 +256,17 @@ def get_archive_prompts(prompt_type: PromptType, user_id: Optional[int]=None, of
     '''
     Get all prompts for archive, including currently active
     '''
-    if (prompt_type == "sprint"):
-        query = "SELECT prompt_id, start, end, rated, active_start, active_end, cmty_added_by, cmty_anonymous, cmty_submitted_time, username FROM sprint_prompts"
+    #if (prompt_type == "sprint"):
+    #    query = "SELECT prompt_id, start, end, rated, active_start, active_end, cmty_added_by, cmty_anonymous, cmty_submitted_time, username FROM sprint_prompts"
     # elif (prompt_type == "marathon")
```
dqian3, line 261: Oh wow, I guess we were thinking about this a while ago/ Also maybe just get rid of these comments.

apis/marathon_api.py
```diff
@@ -122,6 +123,27 @@ def get_all_marathon_prompts():
...
+    ORDER BY prompt_id DESC
+    LIMIT %(p_limit)s
+    """
+
+    db = get_db()
+    with db.cursor(cursor=DictCursor) as cursor:
+        cursor.execute(query, args)
+        results = cursor.fetchall()
+
+        return jsonify(results)
```
dqian3, line 146: Is this just a special case of the archive endpoint? If so you might just want to use  the `prompts.get_archive_prompts` to reduce code duplication? Or are there fields that aren't there?

dqian3 (approved):
Looks good, and seems like you tested it. A few small comments

apis/marathon_api.py
mliu59: Fixed. This was duplicate code

wikispeedruns/prompts.py
mliu59: Removed

PR #563, opened 2024-01-22
Fix Tutorial

dqian3:
Fixed by adding a revision date. This makes it load noticeably slower, but is probably the best option for longevity. Maybe wikipedia will start caching it for us, or we should build our own caching layer.


mliu59:
I tried this out and it was definitely much slower to load.

mliu59 (approved):
approving to allow merge

Issue #463, opened 2022-11-16
Lobby UI/UX polishes


bricehalder:
- [ ] Polishing lobby leaderboards

* leaderboard stats

- [ ] Invite flow is jank

* Passcode be in link like how zoom does it?

- [ ] Lobby page UI is gross especially for lobby admin

- [ ] Editing lobbies

* existing issue: #461 

- [ ] Distinguish between anonymous users w/ same name

PR #569, opened 2024-01-28
add retrievepasscode function to join.js and new button to lobby.html…

Resonantjacket5:
… that adds the passcode to the link, as well as join.js modifying it to fetch the passcode from the url parameters
Screenshot 2024-01-27 at 5 45 52 PM

Copy Link button includes the passcode in the shared url as http://127.0.0.1:5000/lobby/83?passcode=535261

Screenshot 2024-01-27 at 5 46 51 PM

The join page will retrieve the parameter from the urlsearchparameters and then autofill in the text field
(fills it in by when vue is mounted call retrievePasscode function on the join page which checks the url parameter, then modifies the vue.data.passcode field, this then shows up in the passcode ui text field)

### DIFF
--- a/frontend/static/js/pages/lobbys/join.js
+++ b/frontend/static/js/pages/lobbys/join.js
@@ -16,7 +16,26 @@ var app = new Vue({
         message:""
     },
 
+    mounted: function () {
+        this.retrievePasscode();
+    },
+
     methods: {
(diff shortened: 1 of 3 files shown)


Resonantjacket5:
closes #463

mliu59 (approved):
Nice, Thanks for adding this! Tested and looks good on my end

PR #574, opened 2024-01-28
Add small ads to home and play countdown

dqian3:
Adding ads to 2 places that should be viewed a decent amount. Hopefully they're not too intrusive.

Also added a blurb to about explaining them


bricehalder (approved):
$$$

PR #579, opened 2024-01-28
readding "See other" section under title, allows disambig pages

### LINKED ISSUE #506: Articles do not contain "other uses" section at the top
Original Wikipedia article contains links to Pilgrim and Pilgrim (disambiguation) at the top, while the speedruns page does not. It's worth noting that in this particular case the link to Pilgrim is only in the omitted section and therefore unreachable in this path during a speedrun.
image
image

mliu59:
by community request, see issue #506


mliu59:
image

dqian3 (approved):

PR #581, opened 2024-01-28
hints feature

scooterbreak:
hints feature using wiki2vec pathfinding bot

All bot code taken from https://github.com/wikispeedruns/wikispeedruns-bot

### DIFF
--- a/apis/sprints_api.py
+++ b/apis/sprints_api.py
@@ -3,14 +3,22 @@ from flask import Flask, jsonify, request, Blueprint, session
 
 import json
 import datetime
+import wsbot
 
 from app.db import get_db
 from pymysql.cursors import DictCursor
 
 from util.decorators import check_admin, check_request_json
 from wikispeedruns import prompts
+from wsbot.search import GreedySearch, BeamSearch
+from wsbot.embeddings import LocalEmbeddings
+from wsbot.graph import APIGraph, SQLGraph
 
 sprint_api = Blueprint('sprints', __name__, url_prefix='/api/sprints')
+    # this script doesn't work
+    # !./get_embeddings.sh
+embeddings_provider = LocalEmbeddings("data/wiki2vec.txt")
+graph_provider = APIGraph()
 
 
 ### Prompt Management Endpoints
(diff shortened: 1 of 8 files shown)


apis/sprints_api.py
```diff
@@ -3,14 +3,22 @@
...
 
 from util.decorators import check_admin, check_request_json
 from wikispeedruns import prompts
+from wsbot.search import GreedySearch, BeamSearch
+from wsbot.embeddings import LocalEmbeddings
+from wsbot.graph import APIGraph, SQLGraph
 
 sprint_api = Blueprint('sprints', __name__, url_prefix='/api/sprints')
+    # this script doesn't work
+    # !./get_embeddings.sh
```
dqian3, line 19: Yeah, so that is a jupyter notebook specific thing. In a really setting, we would just run this ourselves on the command line somewhere. Maybe modify this comment to say something about needing to download the embeddings.

apis/sprints_api.py
```diff
@@ -168,3 +176,23 @@ def check_duplicate_prompt():
...
+    print(start)
+    print(end)
+
+    if (start is None or end is None): return "Invalid Request", 400
+
+    # which algorithm to use?
+    # greedy = GreedySearch(embeddings_provider, graph_provider)
+    # path = greedy.search(start, end)
+
+    beam = BeamSearch(embeddings_provider, graph_provider)
```
dqian3, line 195: So here, what you should do instead of doing the whole search is do a single step, i.e. look at the pages linked from your current article, and then compare them with the embedding of the goal.

Doing a whole beam search is a lot of unnecessary work.

Note this would require modifying the wsbot library, which I can help with if you're intested. The best way I see of doing that is adding a separate `get_link_rankings` or something to the wsbot lib, that both the `Search` and external users (such as this API) can call.

frontend/templates/play.html
```diff
@@ -57,6 +57,12 @@
                             Current Article<br><strong>[[currentArticle]]</strong>
                         </div>
                     </div>
+                    <div style="float: right;" >
```
dqian3, line 60: This looks good!

dqian3 (commented):
Looks like a great start! There are a few small things that I commented on below that we would need before we merge this in though.

apis/sprints_api.py
scooterbreak: Ah that makes sense I'll look into this. Thanks for reviewing!

scooterbreak:
- Instead of calculating the whole path to the end article, I added a get_next_greedy_link function in search.py that iterates through the current page's links and finds the closest word to the end goal. Added code for greedy search to use the function but left it commented out for now
- Hint text will reset after a new link is clicked

bricehalder (approved):
Looks great! This should surely not crash the server

PR #582, opened 2024-01-28
Added penalty mode functionality

kwonghua:
add a new mode under custom rules when creating a lobby, penalty mode. Adds 20 seconds to the timer when clicking a hyperlink.

### DIFF
--- a/frontend/static/js/pages/play.js
+++ b/frontend/static/js/pages/play.js
@@ -33,6 +33,23 @@ const IS_SCROLL_ON = serverData["scroll"] || null;
 // Get language
 const LANGUAGE = serverData["lang"];
 
+async function getLobby(lobbyId) {
+    const url = `/api/lobbys/${lobbyId}`;
+    const response = await fetch(url);
+
+    if (response.status != 200) {
+        const error = await response.text();
+        alert(error);
+
+        // Prevent are you sure you want to leave prompt
+        window.onbeforeunload = null;
+        window.location.replace("/");   // TODO error page
+        return;
+    }
+
+    return await response.json();
+}
+
 async function getPrompt(promptId, lobbyId=null) {
 
     if(promptId == null){
@@ -106,7 +123,8 @@ let app = new Vue({
         isScroll: null,
 
         anonymous: null,
-        created_username: null
+        created_username: null,
+	    isPenaltyMode: false
     },
 
     mounted: async function() {
@@ -122,6 +140,7 @@ let app = new Vue({
         this.isScroll = IS_SCROLL_ON;
 
         const prompt = await getPrompt(PROMPT_ID, LOBBY_ID);
+	    const lobby = await getLobby(LOBBY_ID);
 
         this.startArticle = prompt["start"];
         this.endArticle = prompt["end"];
@@ -131,6 +150,8 @@ let app = new Vue({
         this.promptPlayed = !!prompt["played"];
         this.promptActive = !!prompt["active"];
         this.promptRated = !!prompt["rated"];
+	
+	    this.isPenaltyMode = !!lobby["rules"]["is_penalty_mode"];
 
         // Use the release date of teh prompt (if it exists) to determine the article revision
         this.revisionDate = prompt?.["active_start"];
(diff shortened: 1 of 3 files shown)


bricehalder:
Demo: 
https://github.com/wikispeedruns/wikipedia-speedruns/assets/17424008/63e1824a-67c5-43a8-a2db-8e8fb21b853d

frontend/static/js/pages/play.js
```diff
@@ -33,6 +33,23 @@ const IS_SCROLL_ON = serverData["scroll"] || null;
 // Get language
 const LANGUAGE = serverData["lang"];
 
+async function getLobby(lobbyId) {
+    const url = `/api/lobbys/${lobbyId}`;
+    const response = await fetch(url);
+
+    if (response.status != 200) {
```
dqian3, line 40: Nice, we should do this more...

frontend/static/js/pages/play.js
```diff
@@ -106,7 +123,8 @@ let app = new Vue({
         isScroll: null,
 
         anonymous: null,
-        created_username: null
+        created_username: null,
+	    isPenaltyMode: false
```
dqian3, line 127: nit: indent

dqian3 (commented):
I like that this adds real time feedback for your clicks in penalty mode, that wasn't something I considered with the leaderboard solution we discussed initially. However, I think ultimately, this should be a combination of this and a change to the leaderboard calculation; this is a good starting point though. 

In fact, the easiest way to do this might be to make a new field for timeElapsed, called timeElapsedWithPenalty (or something less verbose) that the leaderboard can sort by, rather than doing some lengthy SQL calculation. That way we can keep the real time data as well.


Also, it seems like there's mixed indentation, please fix that for later.

frontend/static/js/pages/play.js
```diff
@@ -209,11 +230,17 @@ let app = new Vue({
...
-                this.path.push({
-                    "article": page,
-                    "timeReached": timeElapsed,
-                    "loadTime": loadTimeSeconds
-                });
+		
+		    if (this.isPenaltyMode) {
+		        timeElapsed += this.path.length * 20;
+		        loadTimeSeconds -= 20;
+		    }
```
bricehalder, line 237: This hack is fine for the demo, but for production we need to clean it up. I suggest going with Dan's approach and adding a new field to our schema JSON, maybe `penaltyTime` and we can compute the time on the client by adding `timeElapsed + penaltyTime`.

bricehalder (changes requested):

frontend/static/js/pages/play.js
```diff
@@ -188,7 +210,7 @@ let app = new Vue({
                 this.endTime = Date.now();
             }
 
-            submitRun(PROMPT_ID, LOBBY_ID, this.runId, this.startTime, this.endTime, this.finished, this.path);
+            submitRun(PROMPT_ID, LOBBY_ID, this.runId, this.startTime, this.endTime, this.finished, this.path, this.penaltyTime);
```
bricehalder, line 213: This shouldn't work... I think. Since our path is a JSON object, we can just serialize penaltyTime as part of that without having to modify submitRun. Writing example down below

frontend/static/js/pages/play.js
```diff
@@ -209,6 +231,13 @@ let app = new Vue({
             if (this.path.length == 0 || this.path[this.path.length - 1]["article"] != page) {
                 // Update path
                 let timeElapsed = (Date.now() - this.startTime) / 1000;
+
```
bricehalder, line 234: We could probably do something like 
```
 if (this.isPenaltyMode) {
                    this.penaltyTime += 20;
		        }

                this.path.push({
                    "article": page,
                    "timeReached": timeElapsed,
                    "loadTime": loadTimeSeconds,
                    "penaltyTime": penaltyTime,

                });
```

kwonghua: Okay that makes sense, I’ll look more into that.

kwonghua: Also, for the leaderboard, I made a new option for penalty mode. I was thinking of adding a new column “time (w penalty)” and the penalty mode option for leaderboard will sort by that column. 

Just to clarify is this the approach we’re looking for?

bricehalder: Yeah! I think that's what @dqian3 was talking about

kwonghua:
These are the changes to the leaderboard so far, it sorts by the penalized time which is similar to sorting by length of the shortest path. 

Screen Shot 2024-02-27 at 5 10 31 PM

Also one other concern is that the final click to the end article adds penalized time, not sure if it should be done this way?

bricehalder (approved):
LGTM, maybe make the change to not penalize the end article. Any follow up improvements can be done in later PRs I think

frontend/static/js/pages/play.js
```diff
@@ -209,10 +232,16 @@ let app = new Vue({
             if (this.path.length == 0 || this.path[this.path.length - 1]["article"] != page) {
                 // Update path
                 let timeElapsed = (Date.now() - this.startTime) / 1000;
+		
+		        if (this.isPenaltyMode) {
+                    this.penaltyTime = (this.path.length) * 20;
```
bricehalder, line 237: You can check if `page === this.endArticle` so we don't penalize reaching the goal

Issue #505, opened 2023-06-21
Implement Give Up button on runs


bricehalder:

dqian3:
Closed by #596

PR #596, opened 2024-07-07
Add give up button

mhuangw:
Adds a give up button to return to the home screen during a run. Displays on both collapsed and expanded mode and requires confirmation from the user to prevent misclicks. #505 

Screenshot 2024-07-06 211534


bricehalder (approved):
Nice, this is a highly requested feature!

mliu59:
Nice! As an additional improvement (PR) on top of this, instead of just sending the user to the home screen, I think we can redirect them to the results page and still record their run. See marathon mode's give up button

PR #602, opened 2024-09-03
fixing get_embeddings.sh script

scooterbreak:


dqian3:
Already fixed this in a different commit, oops.

PR #611, opened 2024-10-21
Friend system draft

kwonghua:
draft friend system

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,36 +11,44 @@ This repository holds the code for [wikispeedruns.com](https://wikispeedruns.com
 ## 2. Python Setup
 
 #### Setup Virtual Environment (optional)
+
 We recommend creating a [Python virtual environment](https://docs.python.org/3/tutorial/venv.html)
 for running the server.
+
 ```
 python -m venv env
 ```
 
 For Windows Powershell:
+
 ```
 ./env/Scripts/Activate.ps1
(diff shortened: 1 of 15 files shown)


Resonantjacket5:
For the 
````
CREATE TABLE IF NOT EXISTS `friends` (
    `user_id` INT NOT NULL,
    `friend_id` INT NOT NULL,
    FOREIGN KEY (`user_id`) REFERENCES `users`(`user_id`),
    FOREIGN KEY (`friend_id`) REFERENCES `users`(`user_id`)
);
````

you should add in a primary key to ensure it can query it fast. In mysql this will implicitly add in a clustering index. 

````

CREATE TABLE IF NOT EXISTS `friends` (
    `user_id` INT NOT NULL,
    `friend_id` INT NOT NULL,
    PRIMARY KEY (`user_id`, `friend_id`),
    FOREIGN KEY (`user_id`) REFERENCES `users`(`user_id`),
    FOREIGN KEY (`friend_id`) REFERENCES `users`(`user_id`)
);
````

Note this will allow us to query fast on "user_id" but not on friend_id without a separate secondary index. 


https://stackoverflow.com/questions/3048154/indexes-and-multi-column-primary-keys
https://dev.mysql.com/doc/refman/5.7/en/multiple-column-indexes.html

PR #615, opened 2024-11-09
Added Docker support for dev environment

justanotherinternetguy:
You probably don't want to merge this directly into the master branch for now as it will break the scripts located in `./scripts` for people that aren't using Docker. This is because my code references `/app/config/default.json` (for the docker filesystem) instead of `../config/default.json` (for the normal setup).

Refer to the README in this pr for docker setup instructions.


dqian3:
This looks great! Yeah for the scripts it's probably easy to make a small modification to those scripts so you can pass in those paths. I can help make those changes if you want.

dqian3:
I see you have a draft version of a frontend docker container? Getting that to work would be ideal. I think the ultimate goal is to have the dev just run `docker-compose up` and it will start the flask server and Esbuild script so both can automatically detect changes to code. I'd be happy to get these initial changes in first though and do that in a later PR.

app/app.py
```diff
@@ -14,19 +14,24 @@
...
+    #         static_url_path='/static', 
+    #         static_folder='../frontend/static',
+    #         template_folder='../frontend/templates')
+
     app = Flask(__name__,
-            static_url_path='/static', 
-            static_folder='../frontend/static',
-            template_folder='../frontend/templates')
+        static_url_path='/static', 
+        static_folder='/app/frontend/static',
```
dqian3, line 24: There's two options here: either we can do something using `__file__` to get the relative path, since this won't change. We could also setup some environment variables.

app/db.py
```diff
@@ -10,7 +10,8 @@ def init_app(app):
 def get_conn_info():
     return {
         'user' : current_app.config["MYSQL_USER"],
-        'host' : current_app.config["MYSQL_HOST"],
+        # 'host' : current_app.config["MYSQL_HOST"],
+        'host' : "mysql",
```
dqian3, line 14: I think you can just edit the default `config/default.json` to set this up to work with this docker setup.

frontend/static/js/pages/lobbys/lobby.js
```diff
@@ -231,10 +231,14 @@ var app = new Vue({
         toggleDropdown(id) {
             document.getElementById(id).classList.toggle("active-dropdown")
         },
-
-        async hideLobby(){
+        async hideLobby(event) {
```
dqian3, line 234: If you want to separate this into a separate PR we can merge this stuff in right away.

init.sql
```diff
@@ -0,0 +1,3 @@
+CREATE USER 'user'@'%';
```
dqian3, line 1: Maybe move this to `scripts/`

app/db.py
dqian3: I noticed this happens in a few other places, this also applies.

dqian3 (commented):
Again, looks great! I tried it out myself and it worked like a charm. Like you said yourself, we should fix the scripts so that it isn't specific to Docker, but other than that it's basically ready. I would be happy to make Docker the default instructions for this as well.

PR #617, opened 2024-11-19
Added docker environment

justanotherinternetguy:
Made some small changes, fixed docker filepaths with `__file__`.
LMK if there's anything else :)


README.md
```diff
@@ -4,172 +4,42 @@ This repository holds the code for [wikispeedruns.com](https://wikispeedruns.com
 
 ## 1. Prerequisites
 
-- Python 3.7 or greater
```
dqian3, line 7: Maybe we can move this to a separate ReadMe linked from this main one. I think it's still helpful to have for those who aren't familar with Docker, as well as essentially documentation for what Docker is doing.

dqian3: There are also some optional steps to setup features like achievements and such that could be useful still.

app/app.py
```diff
@@ -14,19 +14,33 @@
...
-            static_folder='../frontend/static',
-            template_folder='../frontend/templates')
+        static_url_path='/static', 
+        static_folder=static_folder,
+        template_folder=template_folder)
 
     app.json_encoder = CustomJSONEncoder
 
-    app.config.from_file('../config/default.json', json.load)
+    app.config.from_file('/app/config/default.json', json.load)
```
dqian3, line 38: Would need to change this here to reference `base_dir` as well so that the local builds don't break.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+services:
+  # MySQL database container
+  mysql:
+    image: mysql:8
+    environment:
+      MYSQL_ROOT_PASSWORD: rootpassword
+      MYSQL_USER: testuser
```
dqian3, line 7: why not just setup the user here?

config/default.json
```diff
@@ -1,7 +1,7 @@
 {
     "SECRET_KEY": "ahhhhhhhhhh",
 
-    "MYSQL_HOST": "127.0.0.1",
+    "MYSQL_HOST": "mysql",
```
dqian3, line 4: This also assumes a Docker setup, which while is the default, doesn't mean we should break other dev environments. I think the best way to handle this would be to have a config/docker.json with this Docker setup, and then set some environment variable in docker or something to signal the application to use that config.

scripts/create_admin_account.py
```diff
@@ -39,10 +41,10 @@ def create_admin_account():
     query = "INSERT INTO `users` (`username`, `hash`, `email`, `email_confirmed`, `admin`, `join_date`) VALUES (%s, %s, %s, %s, %s, %s)"
 
     db = pymysql.connect(
-            user=config["MYSQL_USER"],
-            host=config["MYSQL_HOST"],
-            password=config["MYSQL_PASSWORD"],
-            database=config['DATABASE']
+    user=config["MYSQL_USER"],
+    host="mysql",  # Use the MySQL service name here
```
dqian3, line 45:
```suggestion
    host=config["MYSQL_HOST"],
```

scripts/create_admin_account.py
```diff
@@ -10,12 +9,15 @@
 
 import bcrypt
 
+
+docker_fp = "/app/config/default.json"
```
dqian3, line 13: Same as other comments, this should be redone either using `__file__` or a command line argument to get the path.

dqian3 (commented):
Left some comments marking areas where the docker setup would break local setups. Note that this isn't just so that it's convenient for people who don't want to use Docker, as is these changes would likely break our deployment :).

Also, it seems like the tests (which to be fair are rather outdated) are broken because they reference some `/app` path.

Finally, I might make a few small tweaks to the README myself before merging. If you want, I can also handle all the changes to that (including addressing the comment below about keeping the old instructions). Just lmk!

Thanks again for doing this work!

DeflateAwning:
This would be really helpful to have available for development and bug fixes. Is there anything left to do on this one? @dqian3

Issue #623, opened 2025-04-08
ModuleNotFoundError: No module named 'lib2to3'


JohnnyDHo:
When trying to run the server locally, I get the above error when running this command `flask --app app run --debug` 

This is the last of the error stack
```
  File "/Users/johnnyho/source-code/wiki/wikipedia-speedruns/apis/leaderboard_api.py", line 1, in <module>
    from lib2to3.pgen2.token import OP
ModuleNotFoundError: No module named 'lib2to3'
```

Seems the only mention of `lib2to3` is here https://github.com/wikispeedruns/wikipedia-speedruns/blob/84576f76dea9bf301ba1731fcd26ce7528b4509a/apis/leaderboard_api.py. 

Looks like it was deprecated in Python 3.11 and removed in Python 3.13 which I'm using: https://docs.python.org/3/whatsnew/3.11.html, https://docs.python.org/3/deprecations/pending-removal-in-3.13.html. Is it okay to remove the import since it doesn't seem to be used at a glance in the leaderboard API?

bricehalder:
@JohnnyDHo  Looks like it would be fine to remove

bricehalder:
Closed with #624

PR #624, opened 2025-04-11
[623] Fixed lib2to3 not found error to allow Python 3.13+ to run

JohnnyDHo:
Fixed the above issue: https://github.com/wikispeedruns/wikipedia-speedruns/issues/623

Also ran `npm audit fix` as there were some security vulnerabilities, so the `package-lock.json` was updated. 

Testing: Was able to run the application locally and completed a game, however it seems the `pytest` file is missing so I couldn't run those tests, and I wasn't able to test the leaderboard functionality as it required a public prompt but the ones related to numbers didn't seem completable.

### DIFF
--- a/apis/leaderboard_api.py
+++ b/apis/leaderboard_api.py
@@ -1,4 +1,3 @@
-from lib2to3.pgen2.token import OP
 from flask import Flask, jsonify, request, Blueprint, session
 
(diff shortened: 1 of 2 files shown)


bricehalder (approved):
seems legit

Issue #553, opened 2024-01-15
article previews are unreadable in dark mode


mliu59:
![image](https://github.com/wikispeedruns/wikipedia-speedruns/assets/61032765/3825a846-c940-418f-85f5-08c79e421ff0)

Also see related issue: #537

PR #627, opened 2025-06-12
[553] Updated CSS to use variable background color for 'toolbox' article preview div

JohnnyDHo:
Fix for https://github.com/wikispeedruns/wikipedia-speedruns/issues/553.
Changed to use the variable background color so it will be set correctly regardless of theme. 

Tip for reproducing: Since you can't "inspect" the element because its a popover, you can follow this to be able to inspect: https://stackoverflow.com/questions/17931571/freeze-screen-in-chrome-debugger-devtools-panel-for-popover-inspection

Testing: Ran server locally with changes and confirmed that things worked as expected in both dark mode and light mode.

### DIFF
--- a/frontend/static/stylesheets/play.css
+++ b/frontend/static/stylesheets/play.css
@@ -102,7 +102,7 @@
 .tooltip-container {
     width: 360px;
     font-size: 14px;
-    background: white;
+    background: var(--bs-body-bg);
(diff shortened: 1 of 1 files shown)


JohnnyDHo:
Screenshot 2025-06-12 at 3 00 37 PM
Screenshot 2025-06-12 at 3 01 08 PM
Light and dark mode preview

Issue #632, opened 2025-07-14
Redirect user to home page after login from new password creation


bricehalder:
1/ User does forget password flow
2/ User creates new password
3/ User is logged in, but remains on create password page

Step 3 needs to be changed to redirect to home page after login

anmolsah:
Hi @bricehalder 👋
I’d love to work on this issue! Please assign it to me if it’s still available.
Thanks! 😊

PR #633, opened 2025-07-15
fix: redirect user to home after password reset (#632)

anmolsah:
### Description

This PR addresses [Issue #632](https://github.com/<your-org-or-user>/<repo-name>/issues/632), which describes that after a successful password reset, the user remains on the same page instead of being redirected to the homepage.

### What Has Been Changed

- In the `reset_password.html` template, inside the JavaScript code, the redirect path was updated from:
  
  ```js
  window.location.href = "/login";

### DIFF
--- a/frontend/templates/users/reset_password.html
+++ b/frontend/templates/users/reset_password.html
@@ -45,7 +45,7 @@ async function handleReset(event)
     if (response.status == 200) {
         validationAlert("Password Changed", false);
 
-        window.location.href = "/login";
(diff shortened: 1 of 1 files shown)


bricehalder (changes requested):
Thanks for working on this! This isn't quite right - we still want to refer users to the login page after the password has been reset so they can login with the new password. The behavior to be changed is that after logging in, the user should be redirected home and not back to the reset password page.

PR #634, opened 2025-07-21
Adds redirection of user to home page after login from new password creation

### LINKED ISSUE #632: Redirect user to home page after login from new password creation

nglucas0202:
Closes #632 issue by adding catch for case of referrer being reset page 
image
image
image


bricehalder (approved):
LGTM, will squash the other commits

PR #642, opened 2026-03-28
Update GitHub Actions versions

DeflateAwning:


mliu59 (approved):
verified locally that the pytest bump doesnt break anything else

Issue #641, opened 2026-03-21
Links break mid-game


bycrom27:
While I am playing, randomly none of the links will work. If I click to open in a new tab or window, it gives me a 404. If I click the link, nothing. If I reload, I'm stuck at the starting page with -1 clicks so all I can do is Give Up. That has happened multiple times to me. I am using Chrome on a Chromebook Plus and I have no extensions enabled.

DeflateAwning:
I've experienced this too - giving it a look.

DeflateAwning:
This broke mid-way through the game. It ends up like this. Note the "prevent multiple click" message (in a Chromium-based browser).

<img width="1061" height="150" alt="Image" src="https://github.com/user-attachments/assets/89562a76-9269-40dd-b26b-0208dbf03b7e" />

The first place I searched was `play.js:52` from the error message:

```js
    async function Ao(e, t="en") {
        return await (await fetch(`https://${t}.wikipedia.org/api/rest_v1/page/summary/${e}`, {
            mode: "cors"
        })).json()
    }
```

I don't think that's the issue though. I think the issue primarily spawns from `frontend/static/js/modules/game/articleRenderer.js`:

```js
            // Disable the other linksto prevent multiple clicks
            this.frame.querySelectorAll("a, area").forEach((el) =>{
                el.onclick = (e) => {
                    e.preventDefault();
                    console.log("prevent multiple click");
                };
            });
```

### Reproduction

I figured out how to reproduce it! I'm not certain if this is the only/main way, but it's at least one way.

1. Start a game.
2. Scroll to the bottom of the bage, and click one of the special ISBN book links (they look like "ISBN 9782503524665", for example). 
3. Try to click on any other link to continue playing.
4. In the Dev Tools console, you'll see that you get the "prevent multiple click" message now.

Hopefully this helps someone fix it!

bricehalder:
Thanks for filing this issue! Closed with #648

Issue #549, opened 2024-01-07
Broken anchor tag


mliu59:
https://en.wikipedia.org/wiki/Gerund#Verb_patterns_classified_as_.22gerund.22_use

Reported by cmty, the corresponding anchor tag for this does not work. 

Users have also reported other broken links before, so it could be worth doing a more comprehensive scan for odd links in wikipedia

julianm1138:
Hi, I checked the whole DOM for the page and all the anchor tags looked good. Is there a specific section or heading you are referring to?

mliu59:
Got this error when trying to load "Gerund" with microsoft edge
![image](https://github.com/wikispeedruns/wikipedia-speedruns/assets/61032765/cc9d5ddb-fb82-4650-97d8-7a7b3983ca08)
Although it did not seem to break anchor tags or click events for any other links/sections in the article. 

interestingly, this only occurred on one occasion.

mliu59:
similar issue here: 
![image](https://github.com/wikispeedruns/wikipedia-speedruns/assets/61032765/48c6b03c-c417-4e63-b829-14e58820db96)

on page "Pilgrims (Plymouth colony)"

Issue #605, opened 2024-09-06
Links clicked always -1 and not stopping when hitting goal article


aaosopra:
Hi.

I noticed today that "Links clicked" is always at "-1" and the game does not complete when you have reached the goal article.
The timer stops, but no redirect to result page.

Tested on Microsoft Edge 128.0.2739.54 and Google Chrome Versjon 128.0.6613.119

Issue #551, opened 2024-01-10
All links in article broken


mliu59:
Reported from cmty:
"
Sometimes when I go to a certain page, all the links stop working. Thought it was a subsection issue, but it happens with any link on the page. It almost seems to be page related, because when I retried and did the same route the same issue occurred.
When I went to “association football” that did it. Specific route was Precambrian>Wales>UK>Association Football
"

PR #645, opened 2026-03-28
fix: Clicking certain non-article links caused full game breakage

DeflateAwning:
This should fix issues: #641 and #551. It may also fix Issue #549 and #605 and #481.

Note - Still need to test a bit. Will come back to it eventually. Would appreciate if someone else could give it a try though!


dqian3:
@DeflateAwning Thanks for submitting this! We accidentally ended fixing this on our own PR (#648), but we appreciate the contribution.

DeflateAwning:
I'd take a look at my approach - yours looks over-complicated.

bricehalder:
My PR bundled the fix to the renderer which was the root cause of this issue freezing the game, I also added in the change to exclude non-article links as that was a symptom / simple repro to trigger the issue.

PR #650, opened 2026-04-21
fix: harden Wikipedia fetch failures in gameplay

bricehalder:
### Background

Recent production reports point to a shared failure mode across gameplay, info previews, and quickplay lookups when Wikipedia requests fail or return unexpected responses. The renderer now surfaces article-load failures instead of soft-locking, but transient upstream failures can still leave players staring at a failed load state, and random prompt generation can silently hang when the Wikipedia API returns nothing. This hotfix hardens the shared Wikipedia fetch path while keeping article load time out of sprint timing.

### Change
- Add a shared Wikipedia JSON fetch helper for non-article lookups so transient `429` and `5xx` responses fail more gracefully instead of cascading into broken previews and quickplay helpers.
- Keep article body loads on a single attempt, but pause the sprint timer as soon as an article load starts so the HUD visibly freezes during page loads and resumed timing still excludes load time.
- Normalize preview summary titles and handle missing random-article responses in custom play and lobby prompt generation so those flows show an error instead of hanging.

### Test Plan
`npm run build`

### DIFF
--- a/frontend/static/js/modules/customPlay.js
+++ b/frontend/static/js/modules/customPlay.js
@@ -38,7 +38,14 @@ var CustomPlay = {
             if (this.language === 'en') {
                 [this[prompt]] = await this.$refs.pg.generatePrompt();
             } else {
-                this[prompt] = await getRandomArticle(this.language);
+                const article = await getRandomArticle(this.language);
+                if (!article) {
+                    this.articleCheckMessage = "Unable to fetch a random article right now. Please try again.";
+                    return;
+                }
+
+                this.articleCheckMessage = "";
+                this[prompt] = article;
             }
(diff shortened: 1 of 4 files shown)


mliu59:
@bricehalder were you able to find any links/articles that consistently produce article load errors with 5xx?

The 429 errors seem to be a new introduction to the API just this year: https://www.mediawiki.org/wiki/Wikimedia_APIs/Rate_limits
I believe the way we use the API would put us at a limit of 200 req/min per client browser, which is quite generous. So curious to see if you were able to recreate the 429s consistently. 

Also if we can get "known client" exemptions from wikimedia, that would be pretty sweet :)

PR #651, opened 2026-04-25
Add search functionality for the archive page

nli33:
The main site has 1700+ prompts. This PR allows users to search through the prompts by the "Starting Article" and "Ending Article" fields.

Changes:

- search is done with a simple sql query, no indexes or anything since there are only ~thousands of queries
- relevant search results are slightly shaded in the background and the matching substring is highlighted


mliu59:
Nice, added a fix for wildcard injection in query. @bricehalder requesting a review

bricehalder (approved):

Issue #625, opened 2025-06-12
Anchor tags on homepage doesn't highlight the tab for that part of the page.


JohnnyDHo:
I feel like the UI should highlight the tab like whenever you hover over the link text on the left. See screenshot. 

<img width="918" alt="Image" src="https://github.com/user-attachments/assets/ff81518b-82a4-4e87-aa10-1c7b1dc04de6" />

iriaprados:
Fix on branch issue-#625

PR #654, opened 2026-04-27
Issue #625

iriaprados:
- Linked each menu item to its corresponding section 
- Improved the hover effect for menu options.

### DIFF
--- a/frontend/static/js/modules/game/marathon/marathonPrompts.js
+++ b/frontend/static/js/modules/game/marathon/marathonPrompts.js
@@ -26,7 +26,7 @@ var MarathonPrompts = {
     template: (`
         <div class="card">
             <div class="alert alert-info my-0" role="alert">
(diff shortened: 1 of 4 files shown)


iriaprados:
Hi! I've opened this PR for the issue #625 . Let me know if any changes are needed 🙂

mliu59:
Hello, thanks for submitting the PR! Seems like this does fix #625, but also includes an additional feature for marking the TOC almost like a checklist, which isn't in the original issue. What is the idea behind that?

iriaprados:
Hi, thanks for the review and for pointing this out.

I added a small improvement to the table of contents to improve UX, since it makes it easier to see at a glance which section is selected and where the user currently is in the document. This functionality is non-intrusive and the state is stored only locally, so it does not affect the exported content.

So if you’d prefer to keep this PR focused only on #625, I’m totally fine with removing this part or moving it into a separate PR.

Issue #653, opened 2026-04-26
Penality Mode appears to sort ranking by links instead of penalized time.


defryingpan:
Self-explanatory. To recreate, play a game with Penalty Mode in a party lobby.

Issue #649, opened 2026-04-18
Add custom toggle for portal pages / non-article pages in lobbies, quick play?


bricehalder:
https://discord.com/channels/936202186520678431/936212632866783333/1494981846402863184

Issue #636, opened 2025-09-01
Allow other player in lobby to create prompts


kino90:
Hello! 👋 We’re a small group of friends who love playing Wiki Speedruns.
It would be great if everyone in the lobby could add prompts for others, not just the host.

Would it be possible to support this feature? Thanks a lot! 🙏

Issue #626, opened 2025-06-12
Multiple hosts for private lobbies


bricehalder:
As requested in discord: https://discord.com/channels/936202186520678431/936202187015602218/1382454026989207636

PR #656, opened 2026-05-08
More lobby options

mliu59:
a number of lobby options as mentioned in community requests

- allow multiple hosts in a lobby (not a toggle option, just available for all lobbies). First hosts still have to be designated by original lobby creator #626 
- added option to allow non-host players to submit prompts #636 
- added lobby option to allow all namespace links. Added to quickplay as an option as well #649 
- fixed a leaderboard ranking bug in penalty mode #653

### DIFF
--- a/apis/lobbys_api.py
+++ b/apis/lobbys_api.py
@@ -115,7 +115,10 @@ def add_lobby_prompt(lobby_id):
     user_id = session.get("user_id")
     user_info = lobbys.get_lobby_user_info(lobby_id, user_id)
 
-    if user_info is None or not user_info["owner"]:
+    lobby = lobbys.get_lobby(lobby_id)
+    allow_anyone = lobby and lobby.get("rules", {}).get("allow_anyone_add_prompts", False)
+
+    if user_info is None or (not user_info["owner"] and not allow_anyone):
         return "Only the owner can add prompts to this lobby", 401
 
     start = request.json["start"]
--- a/apis/lobbys_api.py
+++ b/apis/lobbys_api.py
@@ -84,10 +84,19 @@ def get_lobby(lobby_id):
     lobby_user_info = {}
     if "user_id" in session:
         lobby_user_info["name"] = session["username"]
-        lobby_user_info["owner"] = lobbys.get_lobby_user_info(lobby_id, session["user_id"]).get("owner", False)
+        info = lobbys.get_lobby_user_info(lobby_id, session["user_id"]) or {}
+        is_admin = bool(info.get("owner"))
+        is_host = bool(info.get("is_host"))
+        # Keep `owner` for backwards-compatibility with any client that still
+        # reads it; it now means "is admin" (the role formerly known as owner).
+        lobby_user_info["owner"] = is_admin
+        lobby_user_info["is_admin"] = is_admin
+        lobby_user_info["is_host"] = is_host
     else:
         lobby_user_info["name"] = session["lobbys"][str(lobby_id)]
         lobby_user_info["owner"] = False
+        lobby_user_info["is_admin"] = False
+        lobby_user_info["is_host"] = False
 
     lobby_info["user"] = lobby_user_info
 
@@ -101,14 +110,14 @@ def update_lobby(lobby_id):
     user_id = session.get("user_id")
     user_info = lobbys.get_lobby_user_info(lobby_id, user_id)
 
-    if user_info is None or not user_info["owner"]:
-        return "Only the owner can add prompts to this lobby", 401
+    if user_info is None or not user_info["is_host"]:
+        return "Only a host can edit lobby settings", 401
 
     rules = request.json.get("rules")
     name = request.json.get("name")
     desc = request.json.get("desc")
 
-    lobbys.update_lobby(rules=rules, name=name, desc=desc)
+    lobbys.update_lobby(lobby_id, rules=rules, name=name, desc=desc)
 
     return "Updated lobby", 200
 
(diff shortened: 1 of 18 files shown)


wikispeedruns/lobbys.py
```diff
@@ -397,31 +397,45 @@ def get_user_lobbys(user_id: int):
...
+        return cursor.fetchone()["cnt"]
+
 
-    set_host_query = """
+def add_lobby_host(lobby_id: int, target_user_id: int):
+    query = """
     UPDATE user_lobbys
     SET owner=1
     WHERE user_id=%(target_user_id)s AND lobby_id=%(lobby_id)s
     """
```
bricehalder, line 439: There are some implications here like a person becoming an owner of a lobby and then deleting their account will also delete all the original creator's lobbies. We might need to make a new column to track original or primary owner, or add a column for host/admin to differentiate from owner

apis/lobbys_api.py
```diff
@@ -115,7 +115,10 @@ def add_lobby_prompt(lobby_id):
     user_id = session.get("user_id")
     user_info = lobbys.get_lobby_user_info(lobby_id, user_id)
 
-    if user_info is None or not user_info["owner"]:
+    lobby = lobbys.get_lobby(lobby_id)
+    allow_anyone = lobby and lobby.get("rules", {}).get("allow_anyone_add_prompts", False)
+
+    if user_info is None or (not user_info["owner"] and not allow_anyone):
```
bricehalder: [codex] With `allow_anyone_add_prompts=true`, anonymous passcode-joined members see the 'Add Prompt' form in `lobby.js:55-58`. But this route still wears @check_user so they'll always get 401. Should we drop @check_user and accept anon submissions, or gate the UI on `currentUserId !== null`?

frontend/static/js/pages/lobbys/lobby.js
```diff
@@ -42,13 +43,21 @@ var app = new Vue({
...
+        currentUserId: USER_ID,
         users: null,
         anon_users: null,
 
         messageType: null,
         messageTimer: null
     },
 
+    computed: {
+        canAddPrompts() {
```
bricehalder, line 55: [codex] Should `canAddPrompts` also check that we're a logged-in user? Anon (passcode) members see the form but the API rejects them with 401 because `add_lobby_prompt` is @check_user-gated

frontend/static/js/pages/play.js
```diff
@@ -417,7 +422,7 @@ let app = new Vue({
         keydownHandler: function(e) {
             if (this.isFindHotkeyMode) return;
             
-            if ([114, 191, 222].includes(e.keyCode) || ((e.ctrlKey || e.metaKey) && (e.keyCode == 70 || e.keyCode == 71))) {
+            if (["F3", "/", "'"].includes(e.key) || ((e.ctrlKey || e.metaKey) && (e.key == "f" || e.key == "g"))) {
```
bricehalder: nit: Should probably do `e.key.toLowerCase()` for these

frontend/static/js/pages/marathon.js
```diff
@@ -307,7 +307,7 @@ window.onbeforeunload = function() {
 // Disable find hotkeys, players will be given a warning
 window.addEventListener("keydown", function(e) {
     //disable find
-    if ([114, 191, 222].includes(e.keyCode) || ((e.ctrlKey || e.metaKey) && (e.keyCode == 70 || e.keyCode == 71))) {
+    if (["F3", "/", "'"].includes(e.key) || ((e.ctrlKey || e.metaKey) && (e.key == "f" || e.key == "g"))) {
```
bricehalder: Same  e.key.toLowerCase() comment

frontend/static/js/pages/tutorial.js
```diff
@@ -432,7 +432,7 @@ window.addEventListener("beforeunload", beforeUnloadListener);
 // Disable find hotkeys, players will be given a warning
 window.addEventListener("keydown", function(e) {
     //disable find
-    if ([114, 191, 222].includes(e.keyCode) || ((e.ctrlKey || e.metaKey) && (e.keyCode == 70 || e.keyCode == 71))) {
+    if (["F3", "/", "'"].includes(e.key) || ((e.ctrlKey || e.metaKey) && (e.key == "f" || e.key == "g"))) {
```
bricehalder: Same  e.key.toLowerCase() comment

bricehalder (changes requested):
Overall pretty good, just need to address the case where the new lobby host could delete other people's lobbies

frontend/static/js/pages/play.js
mliu59: Addressed in fix commit

apis/lobbys_api.py
mliu59: My original intention for this feature is to allow any new player thats introduced to the game be able to submit prompts without committing a new account. If people want to have only logged in members of their lobby add prompts, they should just add them as cohosts.

mliu59: Added a commit to remove the user gate, but also added a 100 prompt cap for lobbies that have this option enabled

frontend/static/js/pages/lobbys/lobby.js
mliu59: See comments above

wikispeedruns/lobbys.py
mliu59: Addressed this by formally introducing admin vs. host vs. member in commit

a lobby can only have one admin, but can have multiple hosts. Hosts can edit rules/add/remove prompts, but admin is the only one who can manage roles. 

admin can choose to transfer admin to another host. When admin deletes their account, admin role is auto transferred to a host, and if there isnt a host, lobby is deleted.

bricehalder (approved):

